PR #46, opened 2023-02-15
Rename `WaveorderReader` to `ImageReader`

ziw-liu:
This PR does a shallow renaming: the legacy `WaveorderReader` class is now `ImageReader`. Otherwise API remains the same.

### DIFF
--- a/iohub/cli/cli.py
+++ b/iohub/cli/cli.py
@@ -1,7 +1,7 @@
 import click
 
 from iohub._version import __version__
-from iohub.reader import WaveorderReader
+from iohub.reader import ImageReader
 
 VERSION = __version__
 
@@ -19,7 +19,7 @@ def info(files):
     """View basic metadata from a list of FILES"""
     for file in files:
         print(f"Reading file:\t {file}")
-        reader = WaveorderReader(file)
+        reader = ImageReader(file)
         print_reader_info(reader)
 
 
--- a/iohub/pycromanager.py
+++ b/iohub/pycromanager.py
@@ -123,7 +123,7 @@ class PycromanagerReader(ReaderBase):
         Data is not loaded into memory.
 
         Note: The behavior of this function is different from other
-        WaveorderReaders as it return a Dask array rather than a zarr array.
+        ImageReader members as it return a Dask array rather than a zarr array.
 
         # TODO: try casting the dask array into a zarr array
         # using `dask.array.to_zarr()`.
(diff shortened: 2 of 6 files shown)


talonchandler (approved):
Great, thanks @ziw-liu. 

I see that `ImageReader` still uses the (soon to be deprecated) `ZarrReader`. Can I suggest a deprecation warning on `ZarrReader` that points users towards the new `open_ome_zarr`?

AhmetCanSolak (approved):
happy to see this renaming @ziw-liu , I also second the deprecation suggestion from @talonchandler regarding `ZarrReader`

talonchandler:
Thanks for the deprecation warning, @ziw-liu. 

What do you think about reducing the default `log_level` to `logging.WARNING`, too? At least for the `ImageReader`?

I think it's really important to alert users that are currently using `WaveorderReader` (@talonchandler, @edyoshikun , @Soorya19Pradeep) that merely changing `WaveorderReader` to `ImageReader` does not finish the migration to our long term `iohub` solutions. There's more work to do to change to the `ngff` readers.

ziw-liu:
> What do you think about reducing the default `log_level` to `logging.WARNING`, too? At least for the `ImageReader`?

I just realized that I'm not using that custom logger in other parts of the code. So despite being a sub-optimal state the deprecation warning does show up if you invoked it in `__main__` (e.g. a script). We should figure out unified logging too though.

ieivanov (approved):

talonchandler:
> the deprecation warning does show up if you invoked it in __main__ (e.g. a script)

If I run this script: 
```
import logging
from iohub.reader import ImageReader
reader = ImageReader('20200812-CardiomyocyteDifferentiation14-Cycle1.zarr')
print('done')
```
with `python test.py` from a terminal I only see `done`. But if I add `log_level=logging.WARNING` to `ImageReader` then I receive the expected
```
[WARNING: zarrfile:  33 2023-02-15 10:32:08,104] `iohub.zarrfile.ZarrReader` is deprecated and will be removed in the future. For OME-NGFF (OME-Zarr) v0.4 datasets please use `iohub.ngff.open_ome_zarr` instead.
done
```

ziw-liu:
> with `python test.py` from a terminal I only see `done`. But if I add `log_level=logging.WARNING` to `ImageReader` then I receive the expected

Thanks for testing this! I tested in ipython and assumed the same behavior for a script. Then we do need to change log level in this PR.

talonchandler (approved):
Thanks for the fixes @ziw-liu!

PR #50, opened 2023-02-15
Depend on `zarr`

talonchandler:
This PR adds `zarr` to `iohub`'s `install_requires`

A from-scratch install of `iohub` showed that `from iohub.reader import ImageReader` fails since `zarr` isn't an explicit dependency.


talonchandler:
Whoops I messed up this branching...closing and reopening.

PR #51, opened 2023-02-15
Add zarr to install_requires

talonchandler:
This PR adds `zarr` to `iohub`'s `install_requires`.

A from-scratch install of `iohub` showed that `from iohub.reader import ImageReader` fails since `zarr` isn't an explicit dependency.


ziw-liu (approved):
I introduced this regression when moving ome-zarr-py to the dev requirements. Thanks for fixing!

PR #54, opened 2023-02-15
Fix example in README

### LINKED ISSUE #52: `README` example does not run; opens in read-only by default
In the `iohub`'s README we have:

```
with open_ome_zarr("20200812-CardiomyocyteDifferentiation14-Cycle1.zarr") as dataset:
    ...
    new_fov = dataset.create_position("A", "1", "0")  # creates a new fov
    ...
```
which fails because by default `open_ome_zarr` opens in read-only mode. 

Adding `mode='r+'` works, but it modified/added to the dataset I downloaded from the web, then failed on subsequent runs (since the position already existed when I ran it the first time). 

Maybe a more user-friendly README example should read the example dataset, write to a new file, then read+append to that new file?

ziw-liu:
Fix #52

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,11 @@ Load and modify an [official example OME-Zarr](https://zenodo.org/record/7274533
 import numpy as np
 from iohub.ngff import open_ome_zarr
 
-with open_ome_zarr("20200812-CardiomyocyteDifferentiation14-Cycle1.zarr", mode="r", layout="auto") as dataset:
+with open_ome_zarr(
+    "20200812-CardiomyocyteDifferentiation14-Cycle1.zarr",
+    mode="r",
+    layout="auto",
+) as dataset:
     dataset.print_tree()  # prints the hierarchy of the zarr store
     channel_names = dataset.channel_names
     print(channel_names)
@@ -44,15 +48,19 @@ with open_ome_zarr("20200812-CardiomyocyteDifferentiation14-Cycle1.zarr", mode="
     raw_data = img_array.numpy()  # loads a CZYX 4D array into RAM
     print(raw_data.mean())  # does some analysis
 
+# %%
 with open_ome_zarr(
     "max_intensity_projection.zarr",
     mode="w-",
     layout="hcs",
     channel_names=channel_names,
 ) as dataset:
-    dataset.axes = dataset._DEFAULT_AXES[-2:]  # reduce TCZYX axes to only XY
-    new_fov = dataset.create_position("B", "03", "0")  # creates fov with the same path
-    new_fov["0"] = raw_data.max(axis=1)[0]  # max projection along Z axis
+    new_fov = dataset.create_position(
+        "B", "03", "0"
+    )  # creates fov with the same path
+    new_fov["0"] = raw_data.max(axis=1).reshape(
+        (1, 1, 1, *raw_data.shape[2:])
+    )  # max projection along Z axis and prepend dims to 5D
     dataset.print_tree()  # checks that new data has been written
 ```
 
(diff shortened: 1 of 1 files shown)


ziw-liu:
@talonchandler Note that writing to a new store results in a longer example section.

README.md
```diff
@@ -34,13 +34,25 @@ Load and modify an [official example OME-Zarr](https://zenodo.org/record/7274533
...
+    raw_data = img_array.numpy()  # loads a CZYX 4D array into RAM
+    print(raw_data.mean())  # does some analysis
+
+with open_ome_zarr(
+    "max_intensity_projection.zarr",
+    mode="w-",
+    layout="hcs",
+    channel_names=channel_names,
+) as dataset:
+    dataset.axes = dataset._DEFAULT_AXES[-2:]  # reduce TCZYX axes to only XY
```
talonchandler, line 53: Can we skip this and let TCZ have axes lengths 1?

The underscore in `_DEFAULT_AXES` tells me "don't use this" which is a bit inconsistent with showing this in the README example IMO.

README.md
```diff
@@ -34,13 +34,25 @@ Load and modify an [official example OME-Zarr](https://zenodo.org/record/7274533
...
-    first_fov = dataset["B/03/0"]  # lazy Zarr group
-    data = first_fov["0"].numpy()  # loads a CZYX 4D array into RAM
-    print(data.mean())  # does some analysis
-    new_fov = dataset.create_position("A", "1", "0")  # creates a new fov
-    new_fov["0"] = np.ones(data.shape)  # writes some ones to a new Zarr array
+    channel_names = dataset.channel_names
+    print(channel_names)
+    img_array = dataset[
+        "B/03/0/0"
+    ]  # lazy Zarr array for the raw image in the first position
```
talonchandler, line 43: Just formatting, but does this need to be split across multiple lines?

README.md
```diff
@@ -34,13 +34,25 @@ Load and modify an [official example OME-Zarr](https://zenodo.org/record/7274533
 import numpy as np
 from iohub.ngff import open_ome_zarr
 
-with open_ome_zarr("20200812-CardiomyocyteDifferentiation14-Cycle1.zarr") as dataset:
+with open_ome_zarr("test.zarr", mode="r", layout="auto") as dataset:
```
talonchandler, line 37: I actually don't mind the previous idea of having the specific dataset name along with the link directly above it.

talonchandler: I'm open to other solutions, though. Maybe there's a faster way of getting a small `test.zarr` dataset in the user's working directory.

README.md
```diff
@@ -34,13 +34,25 @@ Load and modify an [official example OME-Zarr](https://zenodo.org/record/7274533
...
+    ]  # lazy Zarr array for the raw image in the first position
+    raw_data = img_array.numpy()  # loads a CZYX 4D array into RAM
+    print(raw_data.mean())  # does some analysis
+
+with open_ome_zarr(
+    "max_intensity_projection.zarr",
+    mode="w-",
+    layout="hcs",
+    channel_names=channel_names,
+) as dataset:
```
talonchandler, line 52: Again, just formatting, but it's kind of annoying that it's formatted one way here and another way above.

README.md
ziw-liu: Yes the original file name is better. Reverting

README.md
ziw-liu: This is black behavior. So I guess it's 'consistent' in its own way? I'm open to manually match formatting here though.

README.md
ziw-liu: Good point!

README.md
ziw-liu: Not sure if this helps but now the first one is also line-wrapped.

talonchandler (approved):
Looks good! I'm flexible on formatting.

Issue #45, opened 2023-02-13
Improve API for the NGFF module


ziw-liu:
- [x]  Repurpose the nomenclature from our dependencies:
* Accessing the 5D arrays (zarr arrays) via position['key'] syntax is not intuitive. Let's implement position.data (set or get property) syntax, which mimics napari's syntax. #61

* Rename the ImageArray object to ZarrArray to make it obvious that the object is a zarr array. #60 

- [x] Make it intuitive to read and write tiled acquisitions: #59
* Introduce a `tiles` or `tiled` layout. I picked these words in favor of `multiPos` or `multiFOV`. But, I think either of them is a decent choice. 

- [x] Make it intuitive to over-write channels: #56 #70
*  __setitem__ (position[array_name][:, channel_index] = array) is too complex a usage, let's provide a wrapper `delete_channel` or `overwrite_channel`. `delete_channel` is more broadly useful.

_Originally posted by @mattersoflight in https://github.com/czbiohub/iohub/pull/31#pullrequestreview-1294625315_

PR #61, opened 2023-02-17
Add shorthand to get raw data in NGFF positions

ziw-liu:
Requested by #45.

### DIFF
--- a/iohub/ngff.py
+++ b/iohub/ngff.py
@@ -396,6 +396,34 @@ class Position(NGFFNode):
     def _member_names(self):
         return self.array_keys()
 
+    @property
+    def rawdata(self):
+        """Alias for an array named '0' in the position,
+        which is usually the raw data.
+
+        Returns
+        -------
+        ImageArray
+
+        Raises
+        ------
+        KeyError
+            If no array is named '0'.
+
+        Notes
+        -----
+        Do not depend on this in non-interactive code!
+        The name is hard-coded and is not guaranteed
+        by the OME-NGFF specification.
+        """
+        try:
+            return self["0"]
+        except KeyError:
+            raise KeyError(
+                "There is no array named '0' "
+                f"in the group of: {self.array_keys()}"
+            )
+
     def __getitem__(self, key: Union[int, str]):
         """Get an image array member of the position.
         E.g. Raw-coordinates image, a multi-scale level, or labels
--- a/iohub/ngff.py
+++ b/iohub/ngff.py
@@ -397,9 +397,14 @@ class Position(NGFFNode):
         return self.array_keys()
 
     @property
-    def rawdata(self):
-        """Alias for an array named '0' in the position,
-        which is usually the raw data.
+    def data(self):
+        """.. Warning:
+            This property does *NOT* aim to retrieve all the arrays.
+            And it may also fail to retrive any data if arrays exist but
+            are not named conventionally.
+
+        Alias for an array named '0' in the position,
+        which is usually the raw data (or the finest resolution in a pyramid).
 
         Returns
         -------
--- a/tests/ngff/test_ngff.py
+++ b/tests/ngff/test_ngff.py
@@ -188,6 +188,22 @@ def test_write_ome_zarr(channels_and_random_5d, arr_name):
     channels_and_random_5d=_channels_and_random_5d(),
     arr_name=short_alpha_numeric,
 )
+@settings(max_examples=16, deadline=2000)
+def test_position_rawdata(channels_and_random_5d, arr_name):
+    """Test `iohub.ngff.Position.rawdata`"""
+    channel_names, random_5d = channels_and_random_5d
+    assume(arr_name != "0")
+    with _temp_ome_zarr(random_5d, channel_names, "0") as dataset:
+        assert_array_almost_equal(dataset.rawdata.numpy(), random_5d)
+    with pytest.raises(KeyError):
+        with _temp_ome_zarr(random_5d, channel_names, arr_name) as dataset:
+            _ = dataset.rawdata
+
+
+@given(
+    channels_and_random_5d=_channels_and_random_5d(),
+    arr_name=short_alpha_numeric,
+)
 @settings(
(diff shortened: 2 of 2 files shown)


ziw-liu:
This is a minor feature so I didn't open a dedicated issue. Discussion and suggestions are welcome in the PR (e.g. about naming the attribute).

talonchandler:
Am I understanding this correctly that proposed `dataset.rawdata` only returns the data in the position labelled "0" if it exists?

I think the spirit of my and @mattersoflight's suggestion was to have all of the data available via numerical indexing or via shortcuts. 

Would a `napari`-style API/solution work here? In napari both `viewer.layers['layer-name'].data` and `viewer.layers[0].data` return the same result. Maybe something like `dataset.positions['pos-name'].data` and `dataset.positions[0].data` would be a good equivalent API?

ziw-liu:
> Maybe something like `dataset.positions['pos-name'].data` and `dataset.positions[0].data` would be a good equivalent API?

`dataset.positions['pos-name']` is the same as `dataset['position']` if you meant the path name (e.g. `'A/1/0'`) by `'pos-name'`. However for numeric indexing, I think we have had detailed discussions about how lazy-loading is inherently not compatible with this pattern, and why key-value hierarchy models were chosen for the Zarr-like hierarchical storage protocols. In short, if we want to have position ***lists*** again, it is hard to be compatible with OME-Zarr while also being performant.

ziw-liu:
> I think the spirit of my and @mattersoflight's suggestion was to have all of the data available via numerical indexing or via shortcuts.

For 'having all the data available', consider using `Plate.positions()` generator. With this PR it is possible to do:

```py

with open_ome_zarr("hcs.zarr", mode="r") as dataset:
    for name, pos in dataset.positions():
        print(pos.rawdata.shape)  # now possible
        print(pos['0'].shape)  # same as this
```

mattersoflight:
>
``` 
with open_ome_zarr("hcs.zarr", mode="r") as dataset:
    for name, pos in dataset.positions():
        print(pos.rawdata.shape)  # now possible
        print(pos['0'].shape)  # same as this
 ```
 
 How about an even shorter `pos.data.shape`?

**Comment by ziw-liu** [comment]
> How about an even shorter `pos.data.shape`?

For now `rawdata` corresponds to array `'0'` (by convention the finest resolution level, which is usually raw data). If we add a `Position.data` attribute instead, should that be also just array `'0'`? For reference, napari will return a container of all the arrays with the `data` attribute (loading the [example dataset](https://zenodo.org/record/7274533#.Y-q9uOzMJqv)), and `data[0]` then returns array `'0'`:

```py
>>> viewer.layers['DAPI'].data
<MultiScaleData at 0x178f0d0a0. 5 levels, 'uint16', shapes: ((2, 2160, 5120), (2, 1080, 2560), (2, 540, 1280), (2, 270, 640), (2, 135, 320))>
>>> viewer.layers['DAPI'].data[0]
dask.array<getitem, shape=(2, 2160, 5120), dtype=uint16, chunksize=(2, 2160, 5120), chunktype=numpy.ndarray>
```

I think by not simply naming it as `data`, we can avoid some confusion with the napari behavior, which does not always return raw data.

ziw-liu:
Alternative names I thought of: `basedata` and `image`.

ziw-liu:
```
iohub/multipagetiff.py      162     25    85%
iohub/ngff.py               387     75    81%
iohub/pycromanager.py        69      8    88%
iohub/reader.py             148      8    95%
iohub/reader_base.py         25      4    84%
iohub/singlepagetiff.py     155     20    87%
iohub/upti.py                87     70    20%
iohub/zarrfile.py           136     27    80%
TOTAL                      1388    284    80%
```

ziw-liu:
Thanks to detailed explanation from @mattersoflight, I can now see why the shorthand can be even shorter. The use cases not covered by this new alias is still fully supported, but by KISS the most basic usage is easier to adopt.

@talonchandler @JoOkuma @AhmetCanSolak Now that we are settled with design choices I would appreciate your code review.

iohub/ngff.py
```diff
@@ -396,6 +396,39 @@ def _storage_options(self):
...
+            If no array is named '0'.
+
+        Notes
+        -----
+        Do not depend on this in non-interactive code!
+        The name is hard-coded and is not guaranteed
+        by the OME-NGFF specification.
+        """
+        try:
+            return self["0"]
```
talonchandler, line 425: Could `self["0"]` become `self.array_keys()[0]`? 

Instead of failing if "0" isn't present, this would return the first entry in the `sorted` list of `array_keys()`. This suggestion could lead to confusing behavior if `self.data` returns a different result when you add a new position with a earlier sorted name, but I think I'd prefer this confusing behavior to the confusion of `self.data` returning `KeyError` when I know that there's data in the store.

talonchandler (approved):
One minor question/suggestion...but not blocking. 

Thanks @ziw-liu.

iohub/ngff.py
ziw-liu: My feeling is that we should be very explicit and deterministic about what data the software is reading. IMO the first item of a 'sorted' list is still almost random if the user did not take a look at what files (arrays) they actually have in the folder (group), and if they did, a file name (array key) should be used instead of this shorthand. In short, this attribute should not be used when there *might* be anything other than array `'0'` in the position, and that is (hopefully) emphasized multiple times in the documentation. And I think if `KeyError` is confusing for you, we should aim to improve the error message.

talonchandler: Makes total sense. Thanks 🙏

ziw-liu: > IMO the first item of a 'sorted' list is still almost random

An example would be a position with arrays `['raw', 'projection', 'mask']`, if the list is sorted alphabetically the first item would be 'mask', and is almost certainly not expected, but if they have the same size and dtype this cannot be discovered automatically. This kind of behavior can result in a cryptic errors (better scenario) or false analysis that leaks all the way into a published work (worst case).

Issue #40, opened 2023-02-06
Universal entry points


AhmetCanSolak:
TASKLIST:

- ABC #132 
- OME-Zarr
- OME-TIFF
- ...

<br><br>

>               > At this point I don't think a universal entry point for all the formats is strictly necessary.
> 
> I think that a limited version of a "universal entry point" for internal biohub formats is one the core value propositions of iohub. If we're requiring a user to know their data type and its corresponding API to access data/metadata, then I think adoption of iohub will be limited. 
> 
> The current version of `WaveorderReader` in `waveorder` has been valuable to me because it serves as a universal entry point for compmicro's work. For example, when we started reading pycromanager datasets, @ieivanov added the `PycromanagerReader` class and adapted `WaveorderReader` to read these datasets with the existing API. I could continue to run `WaveorderReader.shape` and `WaveorderReader.get_array` with predictable results, and any unpredictable results were considered bugs and fixed.   
> 
> I understand that there may be difficulty in the details. For example, `PycromanagerReader.get_zarr` might be incorrectly/poorly named, and some dataset types might support different lazy-loading operations (maybe `.get_zarr` is not one of the universal entry points?). But I still think that all of our datasets share some common properties and operations that users will want to perform regularly. I think this list of operations consists of 
> - what is the shape of the data in this dataset? (`.shape`)
> - what is it's basic metadata? (if it's available) (`.channel_names`, `.z_step_size`, `.dtype`, etc)
> - what are the actual values in this dataset? (`.get_array`)
> 
> Can we have a universal entry point that supports these operations?
> 
> _Originally posted by @talonchandler in https://github.com/czbiohub/iohub/issues/31#issuecomment-1419560365_
>             

This deserves a dedicated discussion IMHO, that's why creating a separate issue here.

Moving forward I think we should find ways to engineer a solution that provides a more unified API to the end user. Particularly for adoption of iohub across Biohub, I think this is crucial. Providing list of aforementioned operations(by @talonchandler ) on each dataset is a great start. But I think there is more than that we can do. Like:
- Currently we are implementing readers and writers, but perhaps we can engineer a single dataset class that can handle reading and writing appropriately? and can help us removing redundancies like creating writers from readers and so on.
- We can provide functional APIs like imread/imwrite in `skimage.io` and help scripting users to adopt iohub quicker?
- We can provide a separated metadata API which can enable iohub to be used for inspection before reading different imaging datasets.

[...]

Issue #41, opened 2023-02-07
iohub needs a `convert` CLI command


talonchandler:
iohub needs a simple CLI interface for converting files into the spec-adherent formats that it writes. 

`recOrder` has a simple `convert` command that depends `ZarrConverter` which in turn depends on `WaveorderReader` and `WaveorderWriter`. I think that many parts of the `WaveorderReader` can be reused for iohub's converter, but we have decided not to depend on the `WaveorderWriter`. Therefore, I think the following steps are needed before we're ready for an `iohub convert` PR:

- [ ] We've decided which parts of `WaveorderReader` we're going to reuse
- [ ] the `ZarrConverter` class has been modified for compatibility with the new writer (and possibly modified readers)

@ziw-liu and others, I would appreciate your thoughts on both of these steps.

PR #71, opened 2023-02-24
Image reading entry point

ziw-liu:
This PR removes `ImageReader` (previously `WaveorderReader`) as the extra abstraction layer, and replaces it with a top-level `imread` function that returns a instance of a children of `ReaderBase`. API of the reader object would stay the same, except for the simplification from what would previously be `reader.reader.attr` to `reader.attr`.

Start to address #40.


AhmetCanSolak:
great starting point and it addresses certain aspects of #40 but this will not be enough to close it.

ziw-liu:
> great starting point and it addresses certain aspects of #40 but this will not be enough to close it.

Thanks! I removed #40 from the auto-closing list. This might actually help keep this PR smaller than a full-scale refactor.

mattersoflight:
The `imread` interface is intuitive. I suggest testing this interface via `convert` CLI that converts some example datasets of each type into ome-zarr. I think you have come across all TIFF formats, but not PTI zarr format. You can find multiple example datasets here: https://zenodo.org/record/5951978#.Y_mKt7TMKrM. 

I find that the current `info` CLI has some bugs. I will open an issue.

iohub/reader.py
```diff
@@ -80,7 +81,7 @@ def _check_multipage_tiff(src: str):
     return False
 
 
-def _check_pycromanager(src: str):
+def _check_ndtiff(src: str):
```
talonchandler, line 84: I'm in favor of this renaming. Maybe consider changing `pycromanager` -> `ndtiff` everywhere else for consistency?

iohub/reader.py
```diff
@@ -111,197 +112,70 @@ def _get_sub_dirs(f: str):
...
+        elif _check_multipage_tiff(path):
+            data_type = "ometiff"
+        elif _check_single_page_tiff(path):
+            data_type = "singlepagetiff"
         else:
-            raise NotImplementedError(
-                f"Reader of type {data_type} is not implemented"
+            raise FileNotFoundError(
+                f"No compatible data found under {path}, "
+                "please specify the top level micromanager directory."
```
talonchandler, line 167: Can we improve this message?
https://github.com/mehta-lab/waveorder/issues/93

I think removing the second line will remove much of the confusion.

ziw-liu: The error message can certainly be improved. I think your suggestions in https://github.com/mehta-lab/waveorder/issues/93 are great!

iohub/reader.py
ziw-liu: Will do!

ziw-liu:
> I suggest testing this interface via `convert` CLI that converts some example datasets of each type into ome-zarr.

To keep PRs atomic, I suggest that we implement the convert feature and its CLI option in a separate PR focused on #41.

mattersoflight:
> To keep PRs atomic, I suggest that we implement the convert feature and its CLI option in a separate PR focused on https://github.com/czbiohub/iohub/issues/41.
👍🏼 the entry points can be tested and refined while working on #41.

ziw-liu:
@mattersoflight I took a look at uPTI datasets. They seems to be exactly OME-Zarr stores except for extra calibration arrays in raw datasets (not in processed datasets). So if `imread` is called upon a uPTI store, it can load the positions as arrays with the generic v0.1 OME-Zarr reader. Dispatching to `UPTIReader` however is tricky, in that there is not a consistent pattern in file names or metadata to determine if a store is a uPTI dataset. Can you point me to the specification of this file format so I can learn more?

iohub/reader_base.py
```diff
@@ -1,32 +1,111 @@
...
 
-    def get_array(self, position: int) -> np.ndarray:
-        pass
+    @mm_meta.setter
+    def mm_meta(self, value):
+        if not isinstance(value, dict):
+            raise TypeError(
+                f"Type of `mm_meta` should be `dict`, got `{type(value)}`."
+            )
+        self._mm_meta = value
```
JoOkuma, line 40: Why does the base reader needs `mm_` properties? Shouldn't this be in the `Micromanager` classes?

mattersoflight:
> They seem to be exactly OME-Zarr stores except for extra calibration arrays in raw datasets (not in processed datasets). So if `imread` is called upon a uPTI store, it can load the positions as arrays with the generic v0.1 OME-Zarr reader. 

That's awesome!
>  I had some confusion about what the UPTIReader aims to achieve: I thought it also handles Zarr, but it only handles raw-raw data in TIFF, not the '_raw' datasets in the Zenodo link.

Thanks for that clarification. I'd forgotten how the code evolved. The data acquired with PTI is indeed in TIFF format. 

Examples: `CompMicro/rawdata/hummingbird/LiHao/20210120_A549_RSV_20x` and `CompMicro/rawdata/hummingbird/LiHao/20210226_uPTI_20x_Zebrafish`.

Please test if the current reader can convert this data or not. Great if that works. We can handle any fixes in a separate PR. We also need to update documentation and refactor code in waveOrder to depend on iohub re: PTI.

iohub/multipagetiff.py
```diff
@@ -28,14 +26,12 @@ def __init__(self, folder: str, extract_data: bool = False):
         # Add Initial Checks
         if len(glob.glob(os.path.join(folder, "*.ome.tif"))) == 0:
```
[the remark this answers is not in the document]
ziw-liu: Yes that should be the way. (This code probably predates that standard.)

iohub/reader.py
```diff
@@ -80,14 +81,16 @@ def _check_multipage_tiff(src: str):
...
-def _check_pycromanager(src: str):
+def _check_ndtiff(src: str):
     # go two levels up in case a .tif file is selected
     if src.endswith(".tif"):
         src = os.path.abspath(os.path.join(src, "../.."))
 
     # shortcut, may not be foolproof
     if os.path.exists(os.path.join(src, "Full resolution", "NDTiff.index")):
         return True
+    elif os.path.exists(os.path.join(src, "NDTiff.index")):
```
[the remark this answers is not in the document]
ziw-liu: I would suggest handling this in a separate issue/PR. This legacy pattern might exist elsewhere in the code base too.

iohub/reader.py
```diff
@@ -111,197 +114,70 @@ def _get_sub_dirs(f: str):
...
-        data_type:      (str)
-            whether data is 'ometiff', 'singlepagetiff', or 'zarr'
-        extract_data:   (bool)
-            True if ome_series should be extracted immediately
-        log_level:      (int)
-            One of 0, 10, 20, 30, 40, 50 for
-            NOTSET, DEBUG, INFO, WARNING, ERROR, CRITICAL respectively
-        """
+def imread(
+    path: str,
```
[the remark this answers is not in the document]
ziw-liu: Agreed.

iohub/reader.py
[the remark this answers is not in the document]
JoOkuma: Sounds good!

iohub/reader_base.py
ziw-liu: > How is it saved in the zarr case?

It's a field of zattrs

JoOkuma: What do you think about changing this behavior in this PR or another PR?
In my opinion, the codebase would improve by not having this method in every child, and the functionality could be added with a Mixin, because the `_set_mm_data` looks very similar for every class.

ziw-liu: Both would work. Probably in another PR though so we can merge this one faster. API won't change with a Mixin since currently every child has the method.

ziw-liu:
> Please test if the current reader can convert this data or not.

@mattersoflight Just a reminder that this PR does not feature a converter. That's to be added in #76.

iohub/reader_base.py
JoOkuma: Sounds good!

JoOkuma: Created #85

JoOkuma (approved):

PR #73, opened 2023-02-28
Convenient tiled array I/O with NGFF

### LINKED ISSUE #59: Make it easier to write and read `tiled` dataset.
Taking this point off of #45.

@mattersoflight suggested:
> * Introduce a `tiles` or `tiled` layout. I picked these words in favor of `multiPos` or `multiFOV`. But, I think either of them is a decent choice.

Along with this usage pattern:

https://github.com/czbiohub/iohub/blob/61237f9a9b7a99eb7a1e74ceaa01c0c93204a518/examples/tiles_zarr.py#L1-L41

ziw-liu:
This PR creates an abstraction layer for storing tiled images as chunks in a OME-Zarr array.

Resolves #59.


ziw-liu:
One caveat as of https://github.com/czbiohub/iohub/pull/73/commits/64719d2262dee759f74a0d329682442f4b1b611b: `'auto'` layout in `open_ome_zarr()` does not work since there is no metadata to indicate the tiled layout.

One potential solution is to always return a `TiledPosition` if one of the arrays have `shape[-2:] > chunks[-2:]`. But this can be slower (not by much) and confusing if one does not need `Tiled...` objects.

ziw-liu:
```
iohub/multipagetiff.py      162     25    85%
iohub/ngff.py               459     79    83%
iohub/pycromanager.py        69      7    90%
iohub/reader.py             151      9    94%
iohub/reader_base.py         25      4    84%
iohub/singlepagetiff.py     155     20    87%
iohub/upti.py                87     70    20%
iohub/zarrfile.py           139     29    79%
TOTAL                      1466    290    80%
```

edyoshikun (commented):
This works well for smaller datasets. I had this issue while simulating writing big datasets like the ones we typically acquire (1,6,200,2048,2048), and it wasn't able to write them. It is probably best to do (1,1,1,Y,X) instead of taking `_default_chunking`. 

Otherwise, I was able to write, read, and open in napari without problems.

ziw-liu:
> I had this issue while simulating writing big datasets like the ones we typically acquire (1,6,200,2048,2048), and it wasn't able to write them.

Thanks for testing! The RAM usage skyrockets for large 5D array writing, so lazy writing seems to be broken...

ziw-liu:
@edyoshikun If you do not create large NumPy arrays in RAM but loop over Z it should be fast:
```py
with open_ome_zarr(
    store_path, layout="tiled", mode="a", channel_names=["DAPI", "GFP"]
) as dataset:
    tiles = dataset.make_tiles(
        "tiled_raw", grid_shape=grid_shape, tile_shape=tile_shape
    )
    for row in range(grid_shape[0]):
        for column in range(grid_shape[1]):
            # each tile will be filled with different constant values
            data = np.zeros(shape=tile_shape, dtype=np.uint16) + row + column
            for z in tqdm(range(tile_shape[2])):
                tiles.write_tile(
                    data[:, :, z], row, column, pre_dims=[None, None, z]
                )
```

tests/ngff/test_ngff.py
```diff
@@ -284,6 +285,102 @@ def test_write_more_channels(channels_and_random_5d, arr_name):
...
+    """Test `iohub.ngff.TiledPosition.write_tile()` and `...get_tile()`"""
+    channel_names, random_5d = channels_and_random_5d
+
+    def _tile_data(tiles):
+        for row in range(tiles.rows):
+            for column in range(tiles.columns):
+                yield (
+                    random_5d
+                    / (tiles.rows * tiles.columns + 1)
+                    * (row * column + 1),
```
JoOkuma, line 358: Why is `random_5` divided and multiplied according to the tile coordinates?

JoOkuma: Is it to make it different for each tile?

JoOkuma (commented):
Hi, @ziw-liu .
I left some minor comments where I had a hard time understanding.
Elsewhere, it looks good!

Do you think the `tiled` mode should have the same API to access the positions as `hcs`?
From my point of view, they provide the same (and some extra) functionality with different storage of the image data.

I think the examples should start saving the data at `/tmp` to not pollute home, they could be inside an `if __name__ == "__main__"`. This requires changing the other examples too, so we should it in another PR

ziw-liu:
> Do you think the `tiled` mode should have the same API to access the positions as `hcs`?

I don't think so. `tiled` is essentially just `fov` with a different chunking scheme and some array-level convenience, and the same data can be opened properly with either mode (i.e. with the same pixel values from `__getitem__`). A same API as `hcs` is confusing since they don't write the same data and metadata -- there is no high-content screening hierarchy or metadata involved in `tiled`.

tests/ngff/test_ngff.py
ziw-liu: Yes, so that the test can verify that it got the pixels from the correct tiles.

ziw-liu: I was being lazy here. Could have got the arrays from a hypothesis list strategy wrapping `random5d`, but they should serve the same purpose.

ziw-liu:
> I think the examples should start saving the data at `/tmp` to not pollute home, they could be inside an `if __name__ == "__main__"`. This requires changing the other examples too, so we should it in another PR

Sorry for polluting your `$HOME`... I'm not sure of how to handle `/tmp` on Windows, but that can be a separate issue.

tests/ngff/test_ngff.py
JoOkuma: That's fine; I just wasn't sure it was intentional. Thanks

JoOkuma:
@ziw-liu we could try something like [this](https://stackoverflow.com/a/847874/6748803) for `/tmp`.

JoOkuma (approved):

edyoshikun (approved):
This works well for writing the data! I've tested on smaller datasets and by slicing by Z. 

Thank you @ziw-liu

PR #86, opened 2023-03-10
Makes examples output directory temporary

JoOkuma:
Swap the output directory from $HOME to a temporary directory.

I notice that the examples can be run as notebooks, if we're ok with removing this feature I could add `__name__ == __main__` to this PR.

### DIFF
--- a/examples/multi_fov_hcs_ome_zarr.py
+++ b/examples/multi_fov_hcs_ome_zarr.py
@@ -14,6 +14,7 @@ from iohub.ngff import open_ome_zarr
 # Set storage path
 
 store_path = f"{tempfile.gettempdir()}/hcs.zarr"
+print("Zarr store path", store_path)
 
 # %%
 # Write 5D data to multiple wells.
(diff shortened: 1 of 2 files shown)


JoOkuma:
Hypotheses error, I also get this sometimes when testing `iohub` locally

examples/multi_fov_hcs_ome_zarr.py
```diff
@@ -13,7 +13,7 @@
 # %%
 # Set storage path
 
-store_path = f'{os.path.expanduser("~/")}hcs.zarr'
+store_path = f"{tempfile.gettempdir()}/hcs.zarr"
```
ziw-liu, line 16: Then probably also add a print statement so that it is clear where it got stored? So people can visualize it in napari etc.

ziw-liu:
> Hypotheses error, I also get this sometimes when testing `iohub` locally

Failed health check seems to be just a glitch in stochasticity. Re-running passes. We can suppress that too but maybe not needed in this PR.

ziw-liu (approved):

PR #87, opened 2023-03-11
NGFF method to create zero-filled lazy images

### LINKED ISSUE #64: Chunking option for writing dataset
I was trying to write the array into a newly created position and get an error during write due to size of the`np.array()` which is (1,1,201,2048,2048) `dtype= np.float64`. This error is related to the chunk size, and I realized that we don't have the option to select the desired chunk size of the array and it defaults to the (ZYX shape). 
```python
merged_dataset   = open_ome_zarr(store_path, mode = 'r+')
position = merged_dataset.create_position("position", str(p), "0")
position["0"] = img_normalized
merged_dataset.close()
```
ERROR:
```python
ValueError                                Traceback (most recent call last)
/home/eduardo.hirata/Documents/compmicro-sandbox/zebrafishInfection/ed/20230202_casper_intoto_short1_QLIPP.py in line 3
      179 #%%
      180 position = merged_dataset.create_position("position",str(p),"0")
----> 181 position["0"] = img_normalized

File ~/Documents/iohub/iohub/ngff.py:423, in Position.__setitem__(self, key, value)
    419 if not isinstance(value, np.ndarray):
    420     raise TypeError(
    421         f"Value must be a NumPy array. Got type {type(value)}."
    422     )
--> 423 self.create_image(key, value)

File ~/Documents/iohub/iohub/ngff.py:473, in Position.create_image(self, name, data, chunks, transform, check_shape)
    470 if check_shape:
    471     self._check_shape(data.shape)
    472 img_arr = ImageArray(
--> 473     self._group.array(
    474         name, data, chunks=chunks, **self._storage_options
    475     )
    476 )
    477 self._create_image_meta(img_arr.basename, transform=transform)
    478 return img_arr
...
    120     msg = "Codec does not support buffers of > {} bytes".format(max_buffer_size)
--> 121     raise ValueError(msg)
    123 return arr

ValueError: Codec does not support buffers of > 2147483647 bytes
```

The next thing I tried was to use the `iohub.ngff.Position.create_image()` with `data_chunk = (1, 1, 1, 2048, 2048)` and took a bit to write but it succeded. 
```python
  position = merged_dataset.create_position("position",str(p),"0")
  chunks = (1,1,1,2048,2048)
  position.create_image("0",data=img_normalized, chunks=chunks) 

```

### LINKED ISSUE #84: Writing large data for positions
It is sometimes the case in our [virtual staining pipeline](https://github.com/mehta-lab/microDL) that we will want to write very large arrays to individual positions. The [given example](https://github.com/czbiohub/iohub/blob/main/examples/multi_fov_hcs_ome_zarr.py) for writing positions in an HCS zarr store requires that we hold the whole array in memory at one time:

```py
print(very_large_numpy_array.shape) #array has many time points
#(1000, 3, 25, 2048, 2048)

dataset  = open_ome_zarr(store_path, mode = 'r+')
position = dataset.create_position("position", str(p), "0")
position["0"] = very_large_numpy_array

dataset.close()
```

However, we may only need to grab one slice of size `(1, 1, 1, 2048, 2048)` at a time. Is there a way to build up positions time-step by time-step?
Something like:
```py
print(type(very_large_ImageArray)) #taking advantage of lazy loading
# iohub.ngff.ImageArray

dataset  = open_ome_zarr(store_path, mode = 'r+')
position = dataset.create_position("position", str(p), "0")

position["0"] = some_process(very_large_ImageArray[0])
for timestep in range(1,1000):
   processed_timestep = some_process(very_large_ImageArray[timestep]) #only load one timestep at a time
   position["0"].append_timestep(processed_timestep)

dataset.close()
```

ziw-liu:
Resolves #84.

Fixes #64.

### DIFF
--- a/examples/large_array_ome_zarr.py
+++ b/examples/large_array_ome_zarr.py
@@ -13,7 +13,7 @@ from iohub.ngff import open_ome_zarr
 
 # %%
 # FIXME: set Zarr store path here
-store_path = "../../test-iohub/large.ome.zarr"
+store_path = ""
 
 # %%
 # shape and data type of the large array
--- a/examples/large_array_ome_zarr.py
+++ b/examples/large_array_ome_zarr.py
@@ -1,0 +1,43 @@
+# %%
+# RUNNING THE FOLLOWING CODE WILL WRITE A LARGE FILE
+# This script shows how to store a larger-than-RAM array to a OME-Zarr dataset.
+# The same methods works for HCS datasets too.
+# It can be run as a plain Python script,
+# or as interactive cells in some IDEs.
+
+
+import numpy as np
+from tqdm import tqdm
+
+from iohub.ngff import open_ome_zarr
+
+# %%
+# FIXME: set Zarr store path here
+store_path = "../../test-iohub/large.ome.zarr"
+
+# %%
+# shape and data type of the large array
+# this array is about 10 GB, each time point is about 100 MB
+# it may not be actually larger than RAM but enough for demo
+# monitor the memory usage of python when the following runs
+# and it should take significantly less than 10 GB
+shape = (100, 2, 25, 1024, 1024)
+dtype = np.uint16
+
+
+# %%
+# store this array by looping through the time points
+if store_path:
+    with open_ome_zarr(
+        store_path, layout="fov", mode="w-", channel_names=["DAPI", "GFP"]
+    ) as dataset:
+        img = dataset.create_zeros(
+            name="0",
+            shape=shape,
+            dtype=dtype,
+            chunks=(1, 1, 1, 1024, 1024),  # chunk by XY planes
+        )
+        for t in tqdm(range(shape[0])):
+            # write 4D image data for the time point
+            img[t] = np.ones(shape[1:]) * t
+        dataset.print_tree()
--- a/examples/single_fov_ome_zarr.py
+++ b/examples/single_fov_ome_zarr.py
@@ -8,9 +8,7 @@
 
 import os
 
-import dask.array as da
 import numpy as np
-from tqdm import tqdm
 
 from iohub.ngff import open_ome_zarr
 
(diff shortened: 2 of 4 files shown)


ziw-liu:
```text
iohub/multipagetiff.py      156     25    84%
iohub/ndtiff.py              69     11    84%
iohub/ngff.py               467     83    82%
iohub/reader.py              82     10    88%
iohub/reader_base.py         41      4    90%
iohub/singlepagetiff.py     153     20    87%
iohub/upti.py                86     69    20%
iohub/zarrfile.py           137     29    79%
TOTAL                      1410    298    79%
```

talonchandler:
I just tested this and it's working well. I compared:

- the naive approach of creating a large numpy array and filling the zarr store slice by slice (the numpy array fills memory) to
- the approach in this example where the zarr store is lazily created with zeros, then filled slice by slice (the memory usage never exceeded ~3x the chunk size in my tests)

Great job! Approving. 

My only minor suggestions are to help improve the clarity of the example: 
- if a user isn't familiar with dask, the example might be challenging to understand. Maybe instead of creating a dask array then filling the zarr store with `.compute()` calls, consider filling the zarr store with `np.ones()` calls directly. IMO this clarifies the usage: you don't need dask arrays...the main point is that you can fill a zarr store a slice at a time without filling up memory. 
- consider splitting the `single_fov_ome_zarr.py` script into two or three examples with descriptive names. IMO the current example is getting a bit cluttered and writes multiple zarr stores.

talonchandler (approved):

ziw-liu:
> * if a user isn't familiar with dask, the example might be challenging to understand. Maybe instead of creating a dask array then filling the zarr store with `.compute()` calls, consider filling the zarr store with `np.ones()` calls directly. IMO this clarifies the usage: you don't need dask arrays...the main point is that you can fill a zarr store a slice at a time without filling up memory.

Great point! Fixing.

talonchandler (approved):
LGTM! Thanks for making the changes.

PR #95, opened 2023-03-17
Avoid the `tifffile` version with breaking changes

### LINKED ISSUE #94: New `tifffile` version breaks the OME-TIFF reader
> Tifffile v2023.3.15 includes a new parser for MMStack series and reverts the MicroManager specific optimizations for OME series. Positions are now returned as a dimension in the series rather than as separate series. The dimension order is parsed from the MM metadata and might differ from OME. I only have a limited number of test files, many of which are corrupted in one way or another. Hope it works for you.

@cgohlke Thanks for letting us know! I have yet to change the related code here, so I have to pin the `tifffile` version on the main branch for now to avoid this breaking change.

_Originally posted by @ziw-liu in https://github.com/czbiohub/iohub/issues/66#issuecomment-1472914679_

ziw-liu:
This will be reverted/updated once the OME-TIFF reader gets refactored.

Fixes #94.


edyoshikun (approved):
LGTM

Issue #83, opened 2023-03-08
Automatically generate display metadata for fluorescence channels


ziw-liu:
Currently iohub [detects label-free channels](https://github.com/czbiohub/iohub/blob/c72d46ad3c02ac89f3556383d1ef08461214de3f/iohub/lf_utils.py#L10-L63) and writes OME-NGFF metadata fields to display colors and contrast limits properly. This can be done with common fluorescence channels too. For example GFP is green and DAPI is blue etc.

ziw-liu:
Another possibility is trying to get those from TIFF metadata during conversion. But the metadata (display setting) reading can be quite unstable.

AhmetCanSolak:
this is a great first issue, @ziw-liu  could  you write a detailed task(not  more than a paragraph  probably) on what needs to be done step-by-step and we can try to recruit a new first-time contributor.

ziw-liu:
> this is a great first issue, @ziw-liu could you write a detailed task(not more than a paragraph probably) on what needs to be done step-by-step and we can try to recruit a new first-time contributor.

The [detection function](https://github.com/czbiohub/iohub/blob/c72d46ad3c02ac89f3556383d1ef08461214de3f/iohub/lf_utils.py#L10-L63) is called [here](https://github.com/czbiohub/iohub/blob/c72d46ad3c02ac89f3556383d1ef08461214de3f/iohub/ngff.py#L592-L618). For now the module is called `lf_utils` which stands for 'label-free utilities'. This can be renamed to something more generic like `display_utils`. And the the hard-coded color (`FFFFFF`, RGB hex for white) can become part of the settings dictionary to match channel names. Finally tests can be written to assert that (for example) a channel called 'GFP' has the hex for green in the `omero` color metadata field when it has been written to the OME-Zarr store.

edyoshikun:
@ziw-liu ,  @ieivanov mentioned that Manu's team tends to use `GFP` with grayscale colormap. This is a problem for other datasets where we have multimodal label-free channels (i.e Retardance, States0-4, Phase) and fluorescence in the same store. 
This could be handled by writing different channel names or knowing in advance if the store is just fluorescence channels.

ziw-liu:
> This could be handled by writing different channel names or knowing in advance if the store is just fluorescence channels.

I think iohub should adopt a simple and generic logic. The user can modify the color if necessary. Another possibility is that we implement a convenience method to overwrite all the colormap field in a dataset.

[...]

Issue #74, opened 2023-03-01
add support for writing RGB data


mattersoflight:
We often need to export ND color overlays. @edyoshikun and I ran into this when trying to analyze zebrafish data.
I experimented and found the following:
* RGB data can be written like any other channel (https://github.com/ome/napari-ome-zarr/issues/57).
* `napari-ome-zarr` plugin requires appropriate channel metadata to render the RGB overlays. 

Example FOVs: 
`/Volumes/comp_micro/projects/automation/dataFormats/ome_zarr_demo/test_astro_RGB`

`/hpc/projects/comp_micro/projects/zebrafish-infection/2023_02_02_hummingbird_casper_GFPmacs/hsv_pos7.zarr/positions/0/0/` 

metadata needed in `.zattrs`
```
 "channels": [
            {
                "active": true,
                "color": "FF0000",
                "label": "Red",
                "window": {
                    "end": 0.5,
                    "start": 0.0
                }
            },
            {
                "active": true,
                "color": "00FF00",
                "label": "Green",
                "window": {
                    "end": 0.5,
                    "start": 0.0
                }
            },
            {
                "active": true,
                "color": "0000FF",
                "label": "Blue",
                "window": {
                    "end": 0.5,
                    "start": 0.0
                }
            }
        ],
```

Image appears as follows with napari-ome-zarr
![image](https://user-images.githubusercontent.com/2934183/222041967-58f381ba-91a9-4b98-925a-75a641a4dece.png)

mattersoflight:
@talonchandler mentioned that napari image layer supports the display of RGB or RGBA images if the last dimension of the array has size 3 or 4. 

I see two paths forward:
option 1: `iohub` has the option to add 6D arrays for overlays. These would not conflict with NGFF spec, but allow the custom viewers we write to display.
option 2: We stick with NGFF spec of 5D arrays, but encode each RGB image into 3 channels with names such as `<overlay>_R/G/B`. These layers can be assigned the appropriate color maps and linked.

I like option-2 because it seems that we can use `napari-ome-zarr` out of the box to overlay the color images with other data.

@edyoshikun, @talonchandler , @ziw-liu, others your thoughts?

ziw-liu:
Adding to @edyoshikun's point, I think we should implement NGFF spec to the extent possible in iohub. The custom viewers can also easily read the same data and create one RGB layer when there is a consistent channel naming scheme.

talonchandler:
I think there are two separate items here:

[...]

PR #101, opened 2023-03-29
Adding automatic channel name to Hex color mapping

edyoshikun:
Here I made a dictionary with the most popular colors and most of the channels utilized in our imaging platforms including the label-free channels `(i.e S0,S1,S4, Retardance,Phase,etc)` and fluorophores `(i.e GFP, TXR, DAPI, etc)` to automate the channel name to hex color mapping #83 .

For the color mapping, I have the red fluorophores mapped to magenta as magenta-green is preferred for visualization to be inclusive for folks with deuteranomaly. We currently don't have an easy way to handle HSV #74  except splitting the HSV into RGB, in which each channel is named `red`, `blue`, and `green`.  

Screenshot 2023-03-29 at 2 59 18 PM


examples/single_fov_ome_zarr.py
```diff
@@ -11,22 +11,23 @@
...
 
 from iohub.ngff import open_ome_zarr
+from datetime import datetime
 
 # %%
 # Set storage path
-
-store_path = f"{tempfile.gettempdir()}/ome.zarr"
+timestamp = datetime.now().strftime('%d%H%M%S')
+store_path = '/home/eduardo.hirata/Documents/tmp/' +timestamp+ "ome.zarr"
```
mattersoflight, line 19: We shouldn't hardcode any specific path in the repo. I suggest `os.path.expanduser('~/tmp/...')`.

iohub/display_utils.py
```diff
@@ -0,0 +1,139 @@
...
+    Returns
+    -------
+    str: Hex string corresponding to the RGB value
+    """
+    hex_color = "{:02X}{:02X}{:02X}".format(r, g, b)
+    return hex_color
+
+# Dictionary with key works and most popular fluorescent probes
+channel_colors = {
+    "lime": ["GFP", "Green", "Alexa488", "GCaMP","FITC", "mNeon"],
```
mattersoflight, line 30: green and magenta complement each other, i.e., their overlay is white. Why choose slightly lighter lime?

mattersoflight (commented):
I think #74 requires more detailed discussion and should be handled in a separate PR.

examples/single_fov_ome_zarr.py
edyoshikun: ups didn't mean to push and alter the example code. Ill remove this one. Good catch.

iohub/display_utils.py
edyoshikun: This is also a good point. Initially, I had a typo in my green set as (0,128,0) which made it darker than lime green. Fixed now. Thank you!

iohub/display_utils.py
```diff
@@ -0,0 +1,149 @@
+# Utility functions for displaying data
```
ziw-liu, line 1: Module docstrings should be string literals. See [PEP 257](https://peps.python.org/pep-0257/).

iohub/display_utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+from typing import Tuple
+
+import numpy as np
+
+from iohub.ngff_meta import ChannelMeta, WindowDict
+
+
+def rgb_to_hex(r: int, g: int, b: int) -> str:
```
ziw-liu, line 10: iohub stores color metadata as a pydantic color field which converts different color representations into hex:
https://github.com/czbiohub/iohub/blob/866e6690991fee623a00b8c60a3602a4d7b30922/iohub/ngff_meta.py#L247

So this function is likely redundant.

iohub/display_utils.py
```diff
@@ -0,0 +1,149 @@
...
+    Returns
+    -------
+    str: Hex string corresponding to the RGB value
+    """
+    hex_color = "{:02X}{:02X}{:02X}".format(r, g, b)
+    return hex_color
+
+
+# Dictionary with key works and most popular fluorescent probes
+channel_colors = {
```
ziw-liu, line 31: Can we have this global constant in upper case?

iohub/display_utils.py
```diff
@@ -0,0 +1,149 @@
...
+            clim = channel_settings["Other"]
+            display_color = "FFFFFF"
+
+    # Mapping channel name to color
+    for key in channel_colors:
+        if chan_name in channel_colors[key]:
+            display_color = popular_colors[key]
+            break
+        else:
+            display_color = display_color = "FFFFFF"
```
ziw-liu, line 138:
```suggestion
            display_color = POPULAR_COLORS["white"]
```

iohub/display_utils.py
```diff
@@ -0,0 +1,149 @@
...
+        "Orientation",
+        "State0",
+        "State1",
+        "State2",
+        "State3",
+        "State4",
+    ],
+}
+
+popular_colors = {
```
ziw-liu, line 65:
```suggestion
POPULAR_COLORS = {
```

iohub/display_utils.py
ziw-liu:
```suggestion
CHANNEL_COLORS = {
```

iohub/display_utils.py
```diff
@@ -0,0 +1,149 @@
...
+        "S2": (-0.5, 0.5, -10.0, 10.0),
+        "S3": (-1.0, 1.0, -10.0, -10.0),
+        "Other": (0, U16_FMAX, 0.0, U16_FMAX),
+    }
+    if not clim:
+        if chan_name in channel_settings.keys():
+            clim = channel_settings[chan_name]
+        else:
+            clim = channel_settings["Other"]
+            display_color = "FFFFFF"
```
ziw-liu, line 130:
```suggestion

```

edyoshikun:
@mattersoflight 
One interesting thing to note is that in the from the [CSS3 Specifications](https://www.w3.org/TR/css-color-3/#svg-color) `lime` is (0,255,0) and `green` is just an opaque lime (0,128,0). 

Screenshot 2023-03-30 at 8 47 48 AM

From these suggestions, we realized that the current `pydantic.color.Color()` converts the long hex to short hex (i.e #FFFFFF to #fff). However,  the following function was also not working when modifying the color field and calling `.json()`, so I'm getting rid of it since it doesn't work and it also outputs the wrong hex format.
https://github.com/czbiohub/iohub/blob/866e6690991fee623a00b8c60a3602a4d7b30922/iohub/ngff_meta.py#L254

Instead, we will import the color dictionary from the CSS Specs that already exist in PIL. 

Thank you @ziw-liu!

edyoshikun:
Adding to maybe why the json_encoder didn't work is linked [here](https://stackoverflow.com/questions/73168575/pydantic-color-type-hex-only)

Given that it's used here as well. https://github.com/czbiohub/iohub/blob/866e6690991fee623a00b8c60a3602a4d7b30922/iohub/ngff_meta.py#L303

iohub/display_utils.py
```diff
@@ -1,11 +1,47 @@
-# Utility functions for label-free microscopy data
+""" Utility functions for displaying data """
 
 from typing import Tuple
 
 import numpy as np
+from PIL.ImageColor import colormap
```
ziw-liu, line 6: Maybe also explicitly depend on `pillow` in `setup.cfg`?

ziw-liu:
> Adding to maybe why the json_encoder didn't work is linked [here](https://stackoverflow.com/questions/73168575/pydantic-color-type-hex-only)
> 
> Given that it's used here as well.
> 
> https://github.com/czbiohub/iohub/blob/866e6690991fee623a00b8c60a3602a4d7b30922/iohub/ngff_meta.py#L303

Thanks for pointing this out! I think I fixed the remaining one in https://github.com/czbiohub/iohub/pull/101/commits/411a51350d3d35fdde28e53e021c6697e6995154.

iohub/display_utils.py
edyoshikun: added dependency to current instal version and beyond >= 9.4.0. Is that good?

iohub/display_utils.py
```diff
@@ -50,11 +103,21 @@ def channel_display_settings(
             clim = channel_settings[chan_name]
         else:
             clim = channel_settings["Other"]
+            display_color = color_to_hex("white")
```
ziw-liu, line 106:
```suggestion

```
This line seems to be redundant given [L113-114.](https://github.com/czbiohub/iohub/blob/6dec2b247fe9f3468d638513d3a895fde2100cf4/iohub/display_utils.py#L114)

ziw-liu (approved):
Great first PR!

PR #102, opened 2023-03-30
Use fixed separator `/` to join row_name and col_name

talonchandler:
`os.path.join` uses different separators (`/` or `\`) to join paths, but the NGFF spec only uses `/`. 

The pydantic validator correctly only admits `/`, so it complained when I tried to convert a dataset to ome-zarr v0.4 using the iohub converter on the hummingbird machine. 

I have tested this fix on hummingbird and my personal machine.


ziw-liu (approved):
Thanks!

This really calls for cross-platform testing in CI. I saw this (or the opposite) bug coming in https://github.com/czbiohub/iohub/issues/5#issuecomment-1421169514 but introduced the bug myself anyway...

PR #108, opened 2023-04-14
Added pre-commit

JoOkuma:
Changes:
- added pre-commit
- run pre-commit
- switched python setup pre-req from >3.9 to >=3.10 because pre-commit was failing without it

I would like to include additional changes in this PR, they are styling changes and not necessary, but IMO the code base would improve with it, and they would be run automatically without harming the development.

- [ ] increase line length, I suggest 120, but other values would be fine.
       Most projects use 79 or 80 due to [historical reasons](https://stackoverflow.com/a/111092/6748803), and nowadays, monitors are larger than ever, so increasing the length would make it more appropriate to today's standard (e.g., heavy use of list comprehension) and make part of the code base more readable. For example:

79 max length
```python3
    def get_num_positions(self) -> int:
        return (
            len(self._axes["position"])
            if "position" in self._axes.keys()
            else 1
        )
```
120 max length
```python3
    def get_num_positions(self) -> int:
        return len(self._axes["position"]) if "position" in self._axes.keys() else 1
```
I find the latter more readable.

- [ ]  add pre-commit enforcing a single standard for imports,  https://github.com/MarcoGorelli/absolufy-imports
- [ ]  add pre-commit to update python syntax, https://github.com/asottile/pyupgrade

If others don't agree with the additional suggestions, I'm happy with PR as its.


setup.cfg
```diff
@@ -27,7 +27,7 @@ project_urls =
 [options]
 package = find:
 include_package_data = True
-python_requires = >3.9
+python_requires = >=3.10
```
ziw-liu, line 30: I like personally new Python versions, but for a scientific library, the 3 minor versions we support now was already on the tighter end. Actually, NumPy just dropped 3.8 [today](https://numpy.org/neps/nep-0029-deprecation_policy.html) so we are only starting to be NEP29-compliant.

JoOkuma: Thanks for being this quick.

So this might be a bug because it says > (**greater**) than 3.9.
I'm happy to change to >= (**greater or equal**) than 3.9, it's just that pre-commit was not working with the **greater** symbol 😂 so I had to switch to >= 3.10.

ziw-liu: Bug indeed... Never noticed since local and CI tests has been running fine on 3.9 and the line confusingly has a `=` right before the `>`.

ziw-liu: Actually I will open a separate PR for this

ziw-liu: See #110.

ziw-liu:
@JoOkuma CI for https://github.com/czbiohub/iohub/pull/108/commits/c8c6badc6871011f76ab2de997cd6fab5a81491c is hanging for 1 hr. I'm cancelling it.

ziw-liu:
Oops didn't know 'Fix some PR comment' will close that PR too...

setup.cfg
JoOkuma: Sure, I merged with main.

JoOkuma:
@ziw-liu Are you ok with the other suggestions other than the line length?

ziw-liu:
> @ziw-liu Are you ok with the other suggestions other than the line length?

I think import is already standardized with flake8, black and isort. No need for yet another tool.

For the syntax update thing, I generally don't like anything that 'automagically' changes the semantics of my code... Unless someone can make a strong case of how the lack of such tooling is hurting iohub in major ways I would hesitate to enforce it for every contributor (including myself).

People are of course free to use any tooling locally and code review will hopefully catch any unintended edits.

ziw-liu:
A random thoughr is that the project is not so complicated that we cannot consciously write better code. I recommend keeping the tooling as minimal and predictable as possible.

JoOkuma:
Sounds good! So it should be ok merge.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+    hooks:
+      - id: autoflake
+        args: [--in-place, --remove-all-unused-imports,
+               --ignore-init-module-imports]
+  - repo: https://github.com/PyCQA/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+        args: [--ignore, "E203,W503",
+               --min-python-version, '3.10']
```
ziw-liu, line 30: Can we have 3.9 here?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        args: ["--profile", "black"]
+# syntax linting and formatting
+  - repo: https://github.com/myint/autoflake
+    rev: v1.4
+    hooks:
+      - id: autoflake
```
ziw-liu, line 22: Is this necessary?

docs/source/conf.py
```diff
@@ -3,15 +3,15 @@
 
 import os
 
-# -- General configuration -----------------------------------------------------
+# Add any Sphinx extension module names, as strings. They can be extensions
```
ziw-liu, line 6: @AhmetCanSolak may know more, but I think this file was auto-generated.

.pre-commit-config.yaml
ziw-liu: Or at least provide a way to opt out of this 'feature'...

JoOkuma: it removes unused imports and variables, I don't see any reason why it should be optional, if you don't agree with this I can remove it.

docs/source/conf.py
JoOkuma: Some exceeded the 79-character limit, so it automatically edited some things. I made some changes that it could not fix automatically.

.pre-commit-config.yaml
ziw-liu: Unused imports and variables can sometimes be intentional (although usual due to bad design). I agree that we should not generally allow these, but flake8 should catch those. I think it is the programmer's responsibility to check those issues when the pre-commit runs and the flake8 hook fails, and either fix it (in most cases) or discuss in the PR whether we have to special-case them.

.pre-commit-config.yaml
JoOkuma: Good catch, it was from the previous commit. Fixed.

.pre-commit-config.yaml
ziw-liu: Code quality-wise it's redundant with flake8, and philosophically it delegates the act of consciously writing clean code to a bot that edits code behind the scene. But I'm likely being paranoid here, also curious how others think about this.

ziw-liu:
@JoOkuma Can you also add some instructions to `CONTRIBUTING.md` so everyone knows how to use these?

CONTRIBUTING.md
```diff
@@ -133,7 +133,12 @@ for how this can reveal more bugs.
...
 
-We use [black](https://black.readthedocs.io/en/stable/) to format our code.
-Black installed with `iohub` should automatically use the [settings](./pyproject.toml) in the repository.
+We use [pre-commit](https://pre-commit.com/) to sort imports with [isort](https://github.com/PyCQA/isort), format code with [black](https://black.readthedocs.io/en/stable/), and lint with [flake8](https://github.com/PyCQA/flake8) automatically prior to each commit. To minimize test errors when submitting pull requests, please install pre-commit in your environment as follows:
+
+```bash
+pre-commit install
+```
+
+These packages when executed within `biohub` directory should automatically use the [settings](./pyproject.toml) in the repository.
```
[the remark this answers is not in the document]
ziw-liu: Also there's an `iohub/iohub/` directory. Maybe say 'project root directory' instead?

.pre-commit-config.yaml
keithchev: In spirit, I am with you, but in practice, I think the more layers of defense against messy code, the better.

JoOkuma:
@ziw-liu , any other comment on this PR? last meeting, we decided that duplicating the config, which is minimal, is OK because pre-commit and pyproject do not communicate.

keithchev:
@JoOkuma fwiw I just quickly tested this (on 3.9 with pre-commit 2.20) and it looks like pre-commit hooks do read the config for isort and black from pyproject.toml. 

I can't remember how the conversation went yesterday but it could be that we were confusing the configs for hooks with the config for pre-commit itself, as it seems that pre-commit indeed does not support defining its own config in pyproject.toml.

JoOkuma:
@keithchev Indeed, with this, only the flake8 config is in the pre-commit, since it isn't in the pyproject.toml

ziw-liu (approved):
Looks good, thanks!

keithchev (approved):

PR #110, opened 2023-04-14
Fix python version requirement

ziw-liu:
Fix https://github.com/czbiohub/iohub/pull/108#discussion_r1167040468


edyoshikun (approved):
This LGTM!

PR #112, opened 2023-04-14
Pydata theme for documentation

ziw-liu:
See discussion at #28. We thought TOC trees doesn't work in this theme but it just requires a different syntax.

Also includes minor fixes/rewording of documentation content.

PR #114, opened 2023-04-14
Add Clear Control dataset support

JoOkuma:
This PR implements the [Clear Control](https://github.com/royerlab/opensimview) dataset reader.

Besides a basic working functionality, the CC reader class provides the expected attributes of an array and a least recently used cache, so it **just works** with napari, and no zarr or dask is needed.

For example:

```python3
cc = ClearControlFOV(path, cache=True)
napari.view_image(cc)
napari.run()
```

NOTE: it's loaded just as an array, so there is no scaling according to metadata or splitting the channels into layers.

https://user-images.githubusercontent.com/21022743/232168210-405b4600-9a00-4f43-afab-cb12052b5bff.mp4


examples/view_clearcontrol_dataset.py
```diff
@@ -0,0 +1,39 @@
+import sys
+import tempfile
+import time
+
+from iohub.clearcontrol import (
```
ziw-liu, line 5: Importing a test fixture may be troublesome when we migrate these example scripts to the documentation page.

examples/view_clearcontrol_dataset.py
```diff
@@ -0,0 +1,39 @@
...
+    create_mock_clear_control_dataset,
+)
+
+
+def main():
+    """Example of opening a Clear Control dataset and viewing with napari.
+    """
+    import napari
+
+    if len(sys.argv) < 2:
```
ziw-liu, line 16: Can you document what the arguments should be?

iohub/clearcontrol.py
```diff
@@ -0,0 +1,472 @@
...
+    in_array: np.ndarray,
+    out_path: "StrOrBytesPath",
+    overwrite: bool = False,
+) -> None:
+    """
+    Compresses array and save into output path.
+    This function does not use new functionality from `blosc2` to
+    compress large arrays on purpose, to emulate Clear Control behavior.
+
+    NOTE: function is not optimized and copies the data into bytes,
```
ziw-liu, line 39: Then maybe make this 'private'?

JoOkuma: Done

examples/view_clearcontrol_dataset.py
JoOkuma: Added usage example

JoOkuma:
Thanks @ziw-liu

examples/view_clearcontrol_dataset.py
```diff
@@ -0,0 +1,49 @@
...
+)
+
+
+def main():
+    """Example of opening a Clear Control dataset and viewing with napari.
+
+    Usage
+    -----
+
+    ```
```
ziw-liu, line 17: Markdown syntax may not get rendered properly by sphinx (which uses rst).

Speaking of which, do you want to add this new API to the documentation page?

JoOkuma: Got it, fixed and added it to the API docs

ziw-liu: Sorry my previous comment was wrong/confusing. These should actually either be under `Examples` section and use the doctest syntax ([NumPy docs style guide](https://numpydoc.readthedocs.io/en/latest/format.html#examples)) or be the CLI help message. But not a big deal since we'll need to migrate these to the web page hopefully soon.

ziw-liu (approved):
Code looks good to me! @JoOkuma you can decide whether the other 'public' functions in the module need to be in the documentation.

Didn't test this on any actual data though, but I trust your judgement on when to merge.

JoOkuma:
I'm not sure why Sphinx didn't document them automatically, maybe because there're properties (getters).
I'm fine the way it's.

I merged with main. I added ClearControl first on the documentation to keep it sorted alphabetically, but the list was not fully sorted, let me know if you prefer a different ordering.

ziw-liu:
> I added ClearControl first on the documentation to keep it sorted alphabetically, but the list was not fully sorted, let me know if you prefer a different ordering.

The current ordering is manual, listing the 'primary entry points' on the top. I think it may make sense to move all the one-microscope-specific formats (Dorado, PTI etc.) to a secondary node. But maybe we can deal with that in #112?

Issue #103, opened 2023-04-05
Can `convert` pass the most important information from micromanager image plane metadata?


talonchandler:
**Background:** Earlier version of the tiff-to-zarr `convert` command would write all of the micromanager image plane metadata to text files. Micromanager writes metadata for each plane, so the resulting files were too large to be practical and the latest versions no longer convert this metadata. 

**Proposal:** I think we're now in a position to improve the conversion so that the most important information is correctly read and transferred. I think the most important pieces of image plane metadata are related to the "scale/units" of the data:
- `PixelSizeUm` and `z-step_um`  can be written to the output array scale transformation metadata. 
- Exposure times and time steps might not be suitable for the scale transformations (?), but I think they deserve to live in the converted zarrs
- others? IMO there's not need to transfer all of the device-state metadata, but @ieivanov or @edyoshikun might have preferences here?

talonchandler:
@mattersoflight, @edyoshikun, and I just discussed this issue offline, and we agreed that scale information for the P, T, Z, Y, and X dimensions would be extremely valuable for iohub. 

@ziw-liu, you brought up some technical issues that may make this difficult, and we're hoping to tighten the scope so that this isn't an enormous requirement. I would propose calling this issue "closed" if:

- the converter passes the P, T, Z, Y, and X scale information---no need for any other MM metadata
- for P, we transfer the stage positions from `XPositionUm` and `YPositionUm`
- for T, we transfer `ElapsedTime-ms` (or possibly something like a delta-T from the MDA?)
- for Z, we're already passing `z-step_um`, so I think it's complete. 
- for Y and X, we transfer "PixelSizeUm".
- the passing works for datasets acquired with the most recent version of MM---no need to test against old MM versions. 

I'm happy to tighten this issue even more if there are further difficulties I don't understand.

ziw-liu:
> * the passing works for datasets acquired with the most recent version of MM---no need to test against old MM versions.

In addition to MM version discrepancies, also notice that these fields are specific to OME-TIFF. For example pycro-manager does not have a standard OME metadata tag that stores these information. It will likely require different implementation for each of these fragmented state of TIFF files.

JoOkuma:
I had collaborators ask for the time stamp of the acquisition of each volume.
Can a list of length T be saved somewhere in the metadata after conversion?
Is that information also available when acquired from MM?

ziw-liu:
> Can a list of length T be saved somewhere in the metadata after conversion?

[...]

Issue #66, opened 2023-02-22
iohub's custom OME-TIFF implementation can be slower than tiffile


ziw-liu:
A custom OME-TIFF reader (`iohub.multipagetiff.MicromanagerOmeTiffReader`) was implemented because historically tifffile and AICSImageIO was slow when reading large OME-TIFF series generated by Micro-Manager acquisitions. 

While debugging #65 I found that this implementation does not guarantee data integrity during reading. Before investing more time in fixing it, I think it is worth revisiting the topic of whether it is worth maintaining a custom OME-TIFF reader, given that the more widely adopted solutions have evolved since `waveorder.io`'s designation.

Here is a simple read speed benchmark of tifffile and iohub's custom reader:

![benchmark_plot](https://user-images.githubusercontent.com/67518483/220740761-a4028d70-4976-4c6f-90f9-63f26309e5c4.png)

The test was done on a 123 GB dataset with TCZYX=(8, 9, 3, 81, 2048, 2048) dimensions. Voxels from 2 non-sequential positions was read into RAM in each iteration (N=5).

<details>
  <summary>Test script (click to expand):</summary>

Environment: Python 3.10.8, Linux 4.18 (x86_64, AMD EPYC 7H12@2.6GHz)

  ```py
# %%
import os
from timeit import timeit
import zarr
import pandas as pd

# readers tested
from tifffile import TiffSequence  # 2023.2.3
from iohub.multipagetiff import MicromanagerOmeTiffReader  # 0.1.dev368+g3d62e6f


# %%
# 123 GB total
DATASET = (
    "/hpc/projects/comp_micro/rawdata/hummingbird/Soorya/"
    "2022_06_27_A549cellMembraneStained/"
    "A549_CellMaskDye_Well1_deltz0.25_63X_30s_2framemin/"
    "A549_CellMaskdye_Well1_30s_2framemin_1"
)

POSITIONS = (2, 0)


# %%
def read_tifffile():
    sequence = TiffSequence(os.scandir(DATASET))
    data = zarr.open(sequence.aszarr(), mode="r")
    for p in POSITIONS:
        _ = data[p]
    sequence.close()


# %%
def read_custom():
    reader = MicromanagerOmeTiffReader(DATASET)
    for p in POSITIONS:
        _ = reader.get_array(p)


# %%
def repeat(n=5):
    tf_times = []
    wo_times = []
    for _ in range(n):
        tf_times.append(
            timeit(
[...]

PR #118, opened 2023-04-20
Convert scale metadata

ziw-liu:
Convert pixel and z-step sizes to scale transform in NGFF metadata.

Only implemented for MM OME-TIFF and NDTiff, not for single-page TIFF.

Partial implementation of #103.


ziw-liu:
Converting one of the test datasets:

```log
$ iohub convert -h   
Usage: iohub convert [OPTIONS]

  Converts Micro-Manager TIFF datasets to OME-Zarr

Options:
  -h, --help             Show this message and exit.
  -i, --input DIRECTORY  Input Micro-Manager TIFF dataset directory
                         [required]
  -o, --output PATH      Output zarr store (/**/converted.zarr)  [required]
  -f, --format TEXT      Data type, 'ometiff', 'ndtiff', 'singlepagetiff'
  -g, --grid-layout      Arrange positions in a HCS grid layout
  -p, --label-positions  Dump postion labels in MM metadata to Omero metadata
  -s, --scale-voxels     Write voxel size (XY pixel size and Z-step, in
                         micrometers) as scale coordinate transformation in
                         NGFF

$ iohub convert -i mm2.0-20210713_pm0.13.2_2p_3t_2c_7z_1 -o converted.zarr -f ndtiff --scale-voxels
Dataset opened                 
Status: |████████████████|84/84 (Time Remaining: 00:00), 499.05it/s]
```

Check the written transformation in napari:

napari-metdata showing modified pixel size

ieivanov:
This looks nice! Can `scale_voxels` be the default behavior of the converter?

ziw-liu:
> This looks nice! Can `scale_voxels` be the default behavior of the converter?

I made the CLI flag opt-in so that it's default behavior doesn't change. Also it does not work for all datasets so it might be confusing if it fails to find the metadata and just refuse to convert.

But then I actually made it the default on the python API side... This inconsistency should be fixed and I'm not sure which one is better.

talonchandler:
Agreed, looks very promising! 

I would vote with Ivan for `scale_voxels` being on by default. Maybe `--label-positions` on by default, too?

> Also it does not work for all datasets so it might be confusing if it fails to find the metadata and just refuse to convert.

Maybe throw a warning if the metadata isn't present and continue with the conversion? You could just write ones for the scales if the metadata isn't present.

ziw-liu:
> Maybe throw a warning if the metadata isn't present and continue with the conversion? You could just write ones for the scales if the metadata isn't present.

That's the current implementation.

ziw-liu:
Also a note about why I'm matching OME-XML with some hand-crafted regex and not using [ome-types](https://github.com/tlambert03/ome-types). The data we have off of MM seems to be non-standard/corrupted/outdated (#66) enough that ome-types throws pydantic `ValidationError`s (for wrong types and missing fields) even the XML schema validation is turned off, but also still good enough to extract the metadata we need. This maintenance burden will grow if we keep adding features like this, so just bear in mind...

ziw-liu:
> I would vote with Ivan for `scale_voxels` being on by default. Maybe `--label-positions` on by default, too?

A problem is that it is far more intuitive and easier to document/troubleshoot to add feature flags in CLI commands than it is to turn off a default behavior that fails and throws up a scary stack trace. For example having to do `cp -r` many times is somewhat inconvenient, but having to remember `cp-r-as-default --no-recursive` is probably worse.

edyoshikun:
The HCS grid layout is taken from the metadata too right?

ziw-liu:
> The HCS grid layout is taken from the metadata too right?

Not necessarily. By default it will dump everything into a single row. If the `-p` CLI flag (or `label_positions=True` in Python) is set, the converter will try to guess a grid of **wells** from the stage positions. MM high-content screening metadata is not supported.

iohub/cli/cli.py
```diff
@@ -73,13 +73,22 @@ def info(files, verbose):
     is_flag=True,
     help="Dump postion labels in MM metadata to Omero metadata",
 )
-def convert(input, output, format, grid_layout, label_positions):
+@click.option(
+    "--scale-voxels",
```
talonchandler, line 77: Since we decided to with a default of transferring the scale, maybe this option could be something like "--suppress-scale"?

iohub/convert.py
```diff
@@ -294,6 +299,28 @@ def _get_channel_names(self):
             cns = [str(i) for i in range(self.c)]
         return cns
 
+    def _scale_voxels(self):
+        z_um = self.reader.z_step_size
+        if self.z_dim > 1 and not z_um:
+            logging.warning(
```
talonchandler, line 305: Nice! Makes sense.

iohub/multipagetiff.py
```diff
@@ -292,6 +294,10 @@ def _set_dtype(self):
         tf = TiffFile(self._files[0])
 
         self.dtype = tf.pages[0].dtype
+        xml = tf.ome_metadata
+        # assuming X and Y pixel sizes are the same
+        xy_size = re.search(r"(?<=PhysicalSizeX=\")[\d\.\d]+", xml)
+        self.xy_pixel_size = float(xy_size.group())
```
talonchandler, line 300: Thanks for figuring this out...very valuable.

talonchandler (commented):
I tested this branch on an ome-tiff dataset and the conversion failed (same result with and without `--scale-voxels`). Am I using something incorrectly? Thanks @ziw-liu.

```
iohub convert -i 2022_08_04_recOrder_pytest_20x_04NA/2T_3P_16Z_128Y_256X_Kazansky_1/ -o ./test.zarr
  File "/opt/anaconda3/envs/iohub/bin/iohub", line 8, in <module>
    sys.exit(cli())
  File "/opt/anaconda3/envs/iohub/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
  File "/opt/anaconda3/envs/iohub/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/opt/anaconda3/envs/iohub/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
  File "/opt/anaconda3/envs/iohub/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
  File "/opt/anaconda3/envs/iohub/lib/python3.9/site-packages/click/core.py", line 760, in invoke
  File "/Users/talon.chandler/iohub/iohub/cli/cli.py", line 86, in convert
    converter = TIFFConverter(
  File "/Users/talon.chandler/iohub/iohub/convert.py", line 105, in __init__
    self.reader = read_micromanager(
  File "/Users/talon.chandler/iohub/iohub/reader.py", line 201, in read_micromanager
    return MicromanagerOmeTiffReader(path, extract_data)
  File "/Users/talon.chandler/iohub/iohub/multipagetiff.py", line 53, in __init__
    self._infer_image_meta()
  File "/Users/talon.chandler/iohub/iohub/multipagetiff.py", line 300, in _infer_image_meta
    self.xy_pixel_size = float(xy_size.group())
AttributeError: 'NoneType' object has no attribute 'group'
```

ziw-liu:
Now scaling is the default in CLI.

> I tested this branch on an ome-tiff dataset and the conversion failed (same result with and without `--scale-voxels`). Am I using something incorrectly? Thanks @ziw-liu.

The dataset is missing the metadata field. I couldn't find another field that carries the value, so maybe the microscope was not calibrated to begin with. Now it will convert but log a warning.

iohub/cli/cli.py
ziw-liu: I made it a boolean option that defaults to true.

iohub/multipagetiff.py
```diff
@@ -292,8 +294,19 @@ def _set_dtype(self):
         tf = TiffFile(self._files[0])
 
         self.dtype = tf.pages[0].dtype
+        xml = tf.ome_metadata
+        # assuming X and Y pixel sizes are the same
+        xy_size = re.search(r"(?<=PhysicalSizeX=\")[\d\.\d]+", xml)
```
talonchandler, line 299: Do you know which datasets have `PhysicalSizeX` as a metadata field? This might be a device-specific field. 

I might suggest transferring `PixelSizeUm` instead (see attached screenshot). This field is present in the existing test dataset (though it's set to 0), and it's present if you acquire an MDA with MM's demo camera. 


Screenshot 2023-05-03 at 12 13 05 PM

talonchandler: Also, `PixelSizeUm` uses MM's `> Device > Pixel Size Calibration` feature which is very convenient.

talonchandler (changes requested):
Thanks for fixing the default behavior so that it doesn't fail when metadata is absent. 

I have a suggestion for a better field to transfer from MM...see my comment.

iohub/multipagetiff.py
ziw-liu: > Do you know which datasets have `PhysicalSizeX` as a metadata field?

I got this from the test datasets generated from the demo config.

ziw-liu: OK it's in MM's custom tag. Do note that this tag is also not very stable though.

ziw-liu: Also `0.0` is not a valid scale transform so those datasets will still be set to 1 (identity).

talonchandler: > OK it's in MM's custom tag. Do note that this tag is also not very stable though.

Roger that. I think it's appropriate for us to prioritize micromanager's tags, since that's what we're using internally (and for us OME-TIFF ~= MicroManager multipage TIFF). 

Do you have reason to believe this tag is unstable across MM releases?

> Also 0.0 is not a valid scale transform so those datasets will still be set to 1 (identity).

Great!

talonchandler (approved):

iohub/multipagetiff.py
talonchandler: Great! Thanks a bunch for this PR, @ziw-liu.

PR #119, opened 2023-04-20
add ngff node pyramid initialization

JoOkuma:
Hi, this PR implements the basic functionality to create the multi-scale functionality of an ome zarr dataset.

I'm still figuring out the best way to compute the multi-scale. I can add a helper function in another PR.

My thought is that the multi-scale could be computed with different jobs since IO is the bottleneck, in that case, the code might not leave within iohub.


JoOkuma:
For documentation purpose, my function used to fill the pyramid data.

```python3
def create_time_point_pyramid(t: int, fov: Position, levels: int) -> None:
    import cupy as cp
    from cucim.skimage.transform import downscale_local_mean

    array = cp.asarray(fov["0"][t])
    factors = (2,) * (array.ndim - 1)

    for level in range(1, levels):
        array = cp.stack(
            [downscale_local_mean(array[c], factors) for c in range(array.shape[0])]
        )
        fov[str(level)][t] = array.get()
```

And it's working with napari-ome-zarr.

iohub/pyramid.py
```diff
@@ -0,0 +1,46 @@
+import math as m
```
ziw-liu, line 1: This alias is only used once. Also abbreviating stdlib is not much of a standard practice.

iohub/pyramid.py
```diff
@@ -0,0 +1,46 @@
...
+from copy import deepcopy
+from typing import Sequence, Tuple
+
+from iohub.ngff import Position
+
+
+def _scale_integers(values: Sequence[int], factor: int) -> Tuple[int, ...]:
```
ziw-liu, line 8: https://github.com/czbiohub/iohub/pull/120#discussion_r1174137411

iohub/pyramid.py
```diff
@@ -0,0 +1,46 @@
...
+    """Computes the ceiling of the input sequence divided by the factor."""
+    return tuple(int(m.ceil(v / factor)) for v in values)
+
+
+def initialize_pyramid(fov: Position, levels: int) -> None:
```
ziw-liu, line 13: Can we have this as a method of `Position`?

tests/pyramid/test_pyramid.py
```diff
@@ -0,0 +1,73 @@
...
+            shape=shape,
+            dtype=np.uint16,
+            chunks=(1, 1, 128) + shape[-2:],
+            transform=transform,
+        )
+    
+    return fov
+
+
+def test_pyramid(tmp_path: Path) -> None:
```
ziw-liu, line 43: I think this is a good place to fuzz the test. @JoOkuma Would you mind if I push some changes here?

JoOkuma: Go a head

tests/pyramid/test_pyramid.py
```diff
@@ -0,0 +1,72 @@
...
+
+    fov = open_ome_zarr(ds_path, layout="fov", mode="a", channel_names=channels)
+    transform = [
+        TransformationMeta(
+            type="scale",
+            scale=(1, 1) + scale,
+        )
+    ]
+
+    if "0" not in fov.array_keys():
```
ziw-liu, line 30:
```suggestion
    if "0" not in fov:
```

iohub/ngff.py
```diff
@@ -900,6 +915,41 @@ def update_channel(self, chan_name: str, target: str, data: ArrayLike):
...
+
+        Parameters
+        ----------
+        levels : int
+            Number of down scaling levels, if levels is 1 nothing happens.
+        """
+        array = self.data
+        for level in range(1, levels):
+            factor = 2**level
+            shape = array.shape[:2] + _scale_integers(array.shape[2:], factor)
```
ziw-liu, line 932: An [NGFF image](https://ngff.openmicroscopy.org/0.4/#multiscale-md) may not have all 5 dimensions. See:
https://github.com/czbiohub/iohub/blob/07c7925a735773ab2a652a6177483a7863ef4ea7/iohub/ngff.py#L721-L746

iohub/ngff.py
```diff
@@ -900,6 +907,47 @@ def update_channel(self, chan_name: str, target: str, data: ArrayLike):
...
+        Scales metadata are also updated.
+
+        Parameters
+        ----------
+        levels : int
+            Number of down scaling levels, if levels is 1 nothing happens.
+        """
+        array = self.data
+        for level in range(1, levels):
+            factor = 2**level
```
ziw-liu, line 923: Do we need a check that factor is smaller than shape?

tests/pyramid/test_pyramid.py
```diff
@@ -0,0 +1,84 @@
...
+
+    assert len(fov.array_keys()) == levels
+
+    for level in range(levels):
+        assert str(level) in fov.array_keys()
+
+        level_shape = np.asarray(fov[str(level)].shape)
+        ratio = np.ceil(shape / level_shape).astype(int)
+
+        assert np.all(ratio[:-3] == 1)  # time and channel aren't scaled
```
ziw-liu, line 68: Not sure if this makes any difference, but there's:
https://github.com/czbiohub/iohub/blob/0ddf271002251635bab9ccabab921bc47ef237f0/tests/ngff/test_ngff.py#L15

ziw-liu (approved):
Looks good, thanks!

iohub/ngff.py
JoOkuma: Hi @ziw-liu , I didn't see this comment when merging the PR. We get the ceiling of the `shape / factor`, so it should be fine. Thanks for pointing this out.

tests/pyramid/test_pyramid.py
JoOkuma: I'll refactor it in a different PR in the future.

PR #120, opened 2023-04-20
added helper property to easily access clear control scale

JoOkuma:
Added helper property to easily access Clear Control dataset scale.

This property could be part of the abstract database class. It isn't ergonomic to manually query the scale from the metadata every time.

I don't know why but linting did some changes again.


iohub/clearcontrol.py
```diff
@@ -425,10 +425,20 @@ def metadata(self) -> Dict[str, Any]:
 
         return metadata
 
+    @property
+    def scale(self) -> Tuple[float, ...]:
```
ziw-liu, line 429: `Tuple` is [deprecated](https://docs.python.org/3/library/typing.html#typing.Tuple) for 3.9+.

We still have legacy typing stuff lying around but I think new code should use generics.

ziw-liu: Also is there a reason to use ellipsis if the length is fixed?

JoOkuma: You're right, I removed the `Tuple` usage and converted it to a fixed-length tuple. Thanks!

ziw-liu (approved):
Looks good!

ziw-liu:
Minor comment: can also add a simple test for this.

ziw-liu:
@JoOkuma I'm working on related stuff in #122.

iohub/ngff.py
```diff
@@ -900,6 +900,27 @@ def update_channel(self, chan_name: str, target: str, data: ArrayLike):
...
+        highest resolution scale.
+        """
+        transforms = (
+            self.metadata.multiscales[0].datasets[0].coordinate_transformations
+        )
+        scale_transforms = list(
+            filter(lambda x: x.type == "scale", transforms)
+        )
+
+        if len(scale_transforms) != 1:
```
ziw-liu, line 916: Having multiple items here is not ambiguous. The specification is clear about how they should be applied:

> The transformations in the list are applied sequentially and in order.

So we just need to multiply element-wise.

JoOkuma:
It's ready for review.

iohub/ngff.py
```diff
@@ -900,6 +900,22 @@ def update_channel(self, chan_name: str, target: str, data: ArrayLike):
...
+        Helper function for scale transform metadata of
+        highest resolution scale.
+        """
+        scale = [1] * self.data.ndim
+        transforms = (
+            self.metadata.multiscales[0].datasets[0].coordinate_transformations
+        )
+        for trans in transforms:
+            if trans.type == "scale":
+                assert len(trans.scale) == len(scale)
```
ziw-liu, line 915:
```suggestion
                if len(trans.scale) != len(scale):
                    raise RuntimeError(f"Length of scale transformation {} does not match data dimension {len(scale)}.")
```

JoOkuma: Done

JoOkuma:
@ziw-liu , CI error is not related to the changes.

tests/ngff/test_ngff.py
```diff
@@ -520,3 +521,16 @@ def test_create_position(row, col, pos):
         assert [r["name"] for r in dataset.zattrs["plate"]["rows"]] == [row]
         assert os.path.isdir(os.path.join(store_path, row, col, pos))
         assert dataset[row][col].metadata.images[0].path == pos
+
+
+@given(channels_and_random_5d=_channels_and_random_5d())
+def test_position_scale(channels_and_random_5d):
+    """Test `iohub.ngff.Position.__setitem__()`"""
```
ziw-liu, line 528:
```suggestion
    """Test `iohub.ngff.Position.scale`"""
```

PR #122, opened 2023-04-27
Set transformation metadata for existing NGFF datasets

### LINKED ISSUE #106: modify the coordinateMetadata for old stacks
@ziw-liu @talonchandler @mattersoflight 
From our discussion during our lab meeting, there seems to be a need for us to be able to implement an API call that allows us to modify the coordinate metadata for the datastore.

Currently, to modify the coordinateMetadata I think one has to do the following for each `Position` (please correct me if I'm wrong):
```python
## Test writing to the metadata for existing arrays
dataset = open_ome_zarr(store_path, mode='r+')
print(dataset["0/0/0"].metadata.dict()['multiscales'][0]['coordinate_transformations'])
dataset["0/0/0"].metadata.dict()['multiscales'][0]['coordinate_transformations'][0]['translation']  = [0, 0, 0, 500, 500]
print(dataset["0/0/0"].metadata.dict()['multiscales'][0]['coordinate_transformations'])
dataset.close()
```

This is a lower-priority enhancement, but we will benefit from this to display the multiple positions without having to copy or duplicate data.

ziw-liu:
Resolves #106.

### DIFF
--- a/iohub/ngff.py
+++ b/iohub/ngff.py
@@ -900,6 +900,37 @@ class Position(NGFFNode):
         ortho_sel[ch_ax] = ch_idx
         img.set_orthogonal_selection(tuple(ortho_sel), data)
 
+    def set_transform(
+        self,
+        image: Union[str, Literal["*"]],
+        transform: list[TransformationMeta],
+    ):
+        """Set the coordinate transformations metadata
+        for one image array or the whole FOV.
+
+        Parameters
+        ----------
+        image : Union[str, Literal["*"]]
+            Name of one image array (e.g. "0") to transform,
+            or "*" for the whole FOV
+        transform : list[TransformationMeta]
+            List of transformations to apply
+            (:py:class:`iohub.ngff_meta.TransformationMeta`)
+        """
+        if image == "*":
+            self.metadata.multiscales[0].coordinate_transformations = transform
+        elif image in self:
+            for i, dataset_meta in enumerate(
+                self.metadata.multiscales[0].datasets
+            ):
+                if dataset_meta.path == image:
+                    self.metadata.multiscales[0].datasets[i] = DatasetMeta(
+                        path=image, coordinate_transformations=transform
+                    )
+        else:
+            raise ValueError(f"Key {image} not recognized.")
+        self.dump_meta()
+
 
 class TiledPosition(Position):
     """Variant of the NGFF position node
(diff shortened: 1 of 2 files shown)


ziw-liu:
```text
iohub/__init__.py             4      0   100%
iohub/_version.py             2      0   100%
iohub/clearcontrol.py       179     13    93%
iohub/cli/cli.py             27      0   100%
iohub/convert.py            162     28    83%
iohub/display_utils.py       21      0   100%
iohub/multipagetiff.py      156     25    84%
iohub/ndtiff.py              69      2    97%
iohub/ngff.py               476     62    87%
iohub/ngff_meta.py          181     22    88%
iohub/reader.py             145     15    90%
iohub/reader_base.py         41      4    90%
iohub/singlepagetiff.py     161     19    88%
iohub/upti.py                86     69    20%
iohub/zarrfile.py           137     29    79%
TOTAL                      1847    288    84%
```

JoOkuma:
Hi @ziw-liu, 

My review is below:

 - I noticed you replaced `List` with `list`, I would do the same with `Tuple`.
 - What's the difference between `test_set_transform_image` and `test_set_transform_fov`? They look almost identical, and they have a lot of duplicated code, even though they're part of the tests, it can lead to increased maintenance in the future.
 - The tests in this PR show how it's extremely verbose to interact with metadata / transform in iohub. I think it would benefit from a more friendly API, for example, a `get_transform` method, what do you think @edyoshikun @ziw-liu ?

ziw-liu:
> * The tests in this PR show how it's extremely verbose to interact with metadata / transform in iohub. I think it would benefit from a more friendly API, for example, a `get_transform` method, what do you think @edyoshikun @ziw-liu ?

True... Another possibility we explored during the recent meeting is that we add a transform attribute to `ImageArray`. So FOVs and each resolution level can have their own one-to-one mapped transform attribute.

ziw-liu:
> * What's the difference between `test_set_transform_image` and `test_set_transform_fov`? They look almost identical, and they have a lot of duplicated code, even though they're part of the tests, it can lead to increased maintenance in the future.

One tests setting transformation for an image array (resolution level), the other tests setting transformation for a FOV (all the resolution levels). They are different fields in the metadata.

They do look confusingly similar, but also quite difficult to abstract. I can elevate 3-4 util functions with 1-2 lines each for maintainability's sake but not sure if it's worth it (it will only make the code longer) and whether it will just make it even more confusing.

PR #132, opened 2023-05-18
Universal reader output base classes

JoOkuma:
Hi, this is a work in progress for the base classes so we can advance with the universal reader.

I'm happy to iterate and get feedback from every to have a friendly interface.

I think we could be more restrictive than ome-zarr in some aspects; this would reduce the number of edge cases we need to worry about when implementing our own pipelines (e.g. fixing the shape to 5 and 6)


iohub/dataset_base.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @property
+    @abstractmethod
+    def shape(self) -> tuple[int, int, int, int, int]:
+        # NOTE: suggestion, fix dimension to 5?
+        # We could me more restrictive than ome-zarr
+        raise NotImplementedError
+
+    @abstractmethod
+    def __getitem__(self, key: Any) -> ArrayLike:
```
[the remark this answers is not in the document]
JoOkuma: I was a bit more strict. I added the types supported by zarr. What do you think?

JoOkuma:
> @JoOkuma We don't need `__setitem__` and `__delitem__` since these are read-only.
> 
> Also just to clarify I was talking about `BaseFOVCollection`. So we can do `bool("name" in SomeFOVCollection())`.
> 
> For `BaseFOV` I think it might be ok to ignore pyramids for the time being. (In theory TIFFs can have pyramids too.)

Got it, I updated it accordingly.

iohub/dataset_base.py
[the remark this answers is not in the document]
ziw-liu: What doe ellipsis in `Union` mean?

iohub/dataset_base.py
```diff
@@ -0,0 +1,71 @@
...
+    @property
+    def scale(self) -> list[float]:
+        """Helper function for FOV scale."""
+        raise NotImplementedError
+
+
+class BaseFOVCollection(ABC):
+    @abstractmethod
+    def __contains__(self, position_key: str) -> bool:
+        """Check if position is presented in collection."""
```
ziw-liu, line 56:
```suggestion
        """Check if a position is present in the collection."""
```

iohub/dataset_base.py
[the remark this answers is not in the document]
JoOkuma: The ellipsis is for the tuple, the syntax is a bit misleading.
With python 3.11 it can be typed as, which is clearer in my opinion:

```python3
int | slice | tuple[int | slice, ...]
```

It can be an int, a slice, or a tuple with ints or slices.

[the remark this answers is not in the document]
ziw-liu: Oops wrong bracket count. Thanks for the explanation!

iohub/dataset_base.py
```diff
@@ -0,0 +1,71 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def __getitem__(self, position_key: str) -> BaseFOV:
+        """FOV key position to FOV object."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def __iter__(self) -> Generator[tuple[str, BaseFOV], None, None]:
+        # NOTE is this preferred than the current `.positions` ?
```
ziw-liu, line 70: `NGFFNode` also has `__iter__`, but that only iterates on the immediate next level in the hierarchy (e.g. rows in plate, wells in row etc). The `rows()`, `wells()`, and `positions()` methods allows for iteration on different levels from the same starting point (e.g. plate).

ziw-liu: I think it is still a good idea to always yield `BaseFOV` here. So we will need to change the behavior of the NGFF plate node for compliance.

iohub/dataset_base.py
```diff
@@ -0,0 +1,71 @@
...
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def ndim(self) -> int:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def dtype(self) -> np.dtype:
```
ziw-liu, line 44: You can get `DTypeLike` from `numpy.typing`. So we don't need the np namespace in this file.

iohub/dataset_base.py
```diff
@@ -0,0 +1,71 @@
...
+        # We could me more restrictive than ome-zarr
+        raise NotImplementedError
+
+    @abstractmethod
+    def __getitem__(
+        self,
+        key: Union[int, slice, tuple[Union[int, slice], ...]],
+    ) -> ArrayLike:
+        """
+        Output object must support __array__ interface, np.asarray(...).
```
ziw-liu, line 33:
```suggestion
        Returned object must support the ``__array__`` interface,
        so that ``np.asarray(...)`` will work.
```
Sphinx formatting.

iohub/dataset_base.py
```diff
@@ -0,0 +1,71 @@
...
+    def ndim(self) -> int:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def dtype(self) -> np.dtype:
+        raise NotImplementedError
+
+    @property
+    def scale(self) -> list[float]:
```
ziw-liu, line 48: Should there be a constraint on the length? It's a tricky thing since NGFF has scale in channel dimension, which has no meaning and should always be 1. The time 'scale' is a bit ambiguous too...

ziw-liu: We can also just do `spatial_scale` here and it will always be a 3-tuple.

iohub/dataset_base.py
```diff
@@ -0,0 +1,71 @@
...
+        """Helper function for FOV scale."""
+        raise NotImplementedError
+
+
+class BaseFOVCollection(ABC):
```
ziw-liu, line 53:
```suggestion
class BaseFOVCollection(ABC):
    @abstractmethod
    def __enter__(self) -> BaseFOVCollection:
        "Open the underlying file and return self."
        raise NotImplementedError

    @abstractmethod
    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
        "Close the files."
        raise NotImplementedError
```

Can we also have context manager here?

iohub/dataset_base.py
JoOkuma: `DTypeLike` is a bit broader than `np.dtype`, for example, the string `"unit8"` is `DTypeLike`, if that's OK with you I'm fine with switching.

iohub/dataset_base.py
JoOkuma: Ok, can I leave this to be implemented in a different PR when we inherit the base classes on the `NGFFNode`?

iohub/dataset_base.py
ziw-liu: IMO it's fine since that seems to be how most array libraries (numpy, zarr, torch etc) type their API. But stricter typing doesn't harm so I don't have a strong opinion here.

iohub/dataset_base.py
```diff
@@ -0,0 +1,81 @@
+from abc import ABC, abstractmethod
+from typing import Generator, Union
+
+import numpy as np
+from numpy.typing import ArrayLike
+
+
+class BaseFOV(ABC):
+    # NOTE: not using the `data` method from ngff Position
```
ziw-liu, line 9: Then what is the intended way to access voxel values?

iohub/dataset_base.py
JoOkuma: I'll do `zyx_scale`, it's more informative, IMO.

iohub/dataset_base.py
ziw-liu: Sounds good! Actually I should take care of it once this merges.

iohub/dataset_base.py
ziw-liu: p.s. the `data` property is not exactly 'intended' either...

JoOkuma: I thought of accessing through `__getitem__` and having additional methods to access the multi-scale in ome-zarr datasets. We would have to have a policy to select an array when multiple keys are present.

ziw-liu: In practice most of the acquisition data formats (TIFF, clear control) do not have multi-scale. Should the key always be '0' for them?

ziw-liu: Oh wait did you mean that `__getitem__` directly slices the lowest resolution array? So getting the whole FOV would be `BaseFOV(...)[:]`?

iohub/dataset_base.py
```diff
@@ -0,0 +1,88 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def __getitem__(self, position_key: str) -> BaseFOV:
+        """FOV key position to FOV object."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def __iter__(self) -> Generator[tuple[str, BaseFOV], None, None]:
+        """Iterates over pair o keys and FOV."""
```
ziw-liu, line 87:
```suggestion
        """Iterates over pairs of keys and FOVs."""
```

iohub/dataset_base.py
JoOkuma: I would slice the **highest** resolution.

It would behave sort of like a zarr.
I'm not completely sure if it should load into memory or allow returning any kind of array (e.g. dask array).

I'm more in favor of the latter, and the user should call ```np.asarray``` when it must be loaded into memory, which is not always the case because `numpy` should process any `arraylike` data.

ziw-liu: Got it, so internally it would be like this (for NGFF)?

```py
class Position(...):
    def __getitem__(self, key):
        return self["0"][key]
```

JoOkuma: Yes, we could even allow any key, and use "0" or the first in lexicographic order when more than one key is present.

ziw-liu: I don't think we currently use any other format with pyramids or other data structure that needs multiple array keys in the same FOV. However it does call for more change in the NGFF API since it also does writing, since if `__getitem__` operates on one of the arrays in the FOV, `__setitem__` should match that behavior.

JoOkuma: Yes, it will need to be refactored.

iohub/dataset_base.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def channel_index(self, key: str) -> int:
+        """Return index of given channel."""
+        return self.channels.index(key)
+
+    @property
+    @abstractmethod
+    def shape(self) -> tuple[int, int, int, int, int]:
+        # NOTE: suggestion, fix dimension to 5?
+        # We could me more restrictive than ome-zarr
```
ziw-liu, line 25: If we fix this to a 5-tuple, should `__getitem__` match this behavior? i.e. if an FOV has an underlying array of 3 dimensions, should the slicing be `[0,0,0]` or `[0,0,0,0,0]` to get a scalar?

ziw-liu:
Also we should figure out how the consequent PRs should merge. It's fine if this one merges into main but if we keep the refactoring of readers atomic and distributed, asynchronously merging into main will result in a broken default branch.

Maybe make a new target branch (e.g. `uapi-readers`) once this is merged?

iohub/dataset_base.py
keithchev: From the perspective of non-expert end users, I think it would be confusing if the pixel values in a 3D image had to be accessed using something like `[x, y, z, 0, 0]`. (It would be even worse if the placeholder channel and time dimensions came first, because then a naive indexing using `[x, y, z]` would not only not result in a scalar but would be entirely incorrect)

JoOkuma:
> Also we should figure out how the consequent PRs should merge. It's fine if this one merges into main but if we keep the refactoring of readers atomic and distributed, asynchronously merging into main will result in a broken default branch.
> 
> Maybe make a new target branch (e.g. `uapi-readers`) once this is merged?

I like the idea of having a separate branch.

I'm happy to take a few days next week and be done with it. I'm free after next Wednesday, but I'll need help with `tiff` because I'm not super familiar with it.

iohub/dataset_base.py
keithchev: Fair enough!

ziw-liu: There is a glitch in slicing with all 5 dimensions though. Translating the slicing makes it impossible to return a lazy Zarr array, removing many useful attributes (append, chunks, fancy slicing etc) that we currently inherit from `zarr.Array`. And at writing time, a NumPy array is directly written to Zarr, preserving shape, and it would be very confusing to have to slice them differently (or even having two different array types depending on how the file is opened).

JoOkuma: Uhm... an alternative is to implement a method on `BaseFOV` that translate the 5-dim indexing to the original array, and when requesting the whole array we can call `np.expand_dims`.

ziw-liu: > Uhm... an alternative is to implement a method on `BaseFOV` that translate the 5-dim indexing to the original array, and when requesting the whole array we can call `np.expand_dims`.

Yeah I think this is reasonable. It actually has to call `expand_dims` even when not requesting the whole array. So we might as well just delegate this to dask or tensorstore by constructing the reshaped always-lazy da/ts array with `dask()` or `tensorstore()` as we currently have with `ngff.ImageArray.numpy()`.

ziw-liu: > So we might as well just delegate this to dask or tensorstore by constructing the reshaped always-lazy da/ts array with `dask()` or `tensorstore()` as we currently have with `ngff.ImageArray.numpy()`.

This way we can also wrap dask in xarray and delegate the dimension naming API too.

edyoshikun (commented):
This all LGTM! Thank you guys.  I like the `BaseCollectionFOV` as an iterable that we can use to access individual positions. 

A couple of points just to make sure this will behave as intended. 
- For this universal reader, we will call `BaseFOV[0,0,0]` this will  return a (1,1,1,Z,Y,X) 5D array?

ziw-liu:
> * For this universal reader, we will call `BaseFOV[0,0,0]` this will  return a (1,1,1,Z,Y,X) 5D array?

My understanding is that it will behave as normal array slicing does. So ZYX (3D) in this case.

iohub/dataset_base.py
```diff
@@ -107,37 +108,69 @@ def __eq__(self, other: BaseFOV) -> bool:
...
-    def __exit__(
+    def __init__(
         self,
-        exc_type: Optional[Type[BaseException]],
-        exc_val: Optional[BaseException],
-        exc_tb: Optional[TracebackType],
-    ) -> bool:
-        """Close the files."""
-        raise NotImplementedError
+        data_dict: Optional[dict[str, BaseFOV]] = None,
```
ziw-liu, line 118: This constructor needs reference to all of the FOVs at creation time, meaning that for HCS OME-Zarr the entire tree has to be walked upon reading the plate. Is there a way to do this lazily as it is currently?

iohub/dataset_base.py
ziw-liu: @JoOkuma and I had a chat and we will try out zarr's `view` for this.

iohub/dataset_base.py
JoOkuma: HCS OME-Zarr could have its own constructor and provide no parameter do the parent `FOVCollection` class, or do you prefer an abstract base class that doesn't initialize anything?

ziw-liu: But then the attributes initialized are expected elsewhere, which makes it difficult to maintain. I think the base class should be abstract and the implementation here can become something like `FOVDict`

iohub/dataset_base.py
```diff
@@ -0,0 +1,176 @@
...
+    _root: Path
+
+    @property
+    @abstractmethod
+    def axes_names(self) -> list[str]:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def channels(self) -> list[str]:
```
talonchandler, line 25: Maybe `channels` -> `channel_names`, since `channels` might make you think you're getting the number of channels?

iohub/dataset_base.py
```diff
@@ -0,0 +1,176 @@
...
+    ) -> ArrayLike:
+        """
+        Returned object must support the ``__array__`` interface,
+        so that ``np.asarray(...)`` will work.
+        """
+        raise NotImplementedError
+
+    @property
+    def ndim(self) -> int:
+        return 5
```
talonchandler, line 93: I like this choice.

iohub/dataset_base.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @property
+    @abstractmethod
+    def dtype(self) -> np.dtype:
+        raise NotImplementedError
+
+    @property
```
talonchandler, line 100: `BaseFOV` has `zyx_scale` and `channels/channel_names`, but no property related to the time dimension. 

Maybe `def t_scale(self) -> float`? Or maybe we want to support irregular time steps with `def times(self) -> list[float]`?

iohub/dataset_base.py
talonchandler, line 25: Is there a "correct" way to enforce that `len(BaseFOV.channels)` should be identical to `BaseFOV.shape[1]` within the abstract base class?

Similarly, if we include a `BaseFOV.times` then `len(BaseFOV.times)` should be identical to `BaseFOV.shape[0]`.

talonchandler (approved):
Looks great to me! I'm excited. 

Thanks for pushing on this @JoOkuma.

iohub/dataset_base.py
JoOkuma: Do you this the `list[float]` should be enforced by the API, or is it very specific to some datasets?

iohub/dataset_base.py
JoOkuma: I'm not really sure what's the best practice. I think that should be covered by tests, another option is to add an assert in this class but it will only be useful if `super().channels` is called.

iohub/dataset_base.py
JoOkuma: @ziw-liu @talonchandler What do you think about ditching `zyx_scale` and having `scale` that returns a 5-tuple of floats?

JoOkuma:
@ziw-liu, I renamed files from `database_base.py` to `fov.py`.

iohub/fov.py
```diff
@@ -0,0 +1,215 @@
...
+                f"Found {key} with type {type(key)}"
+            )
+
+        if not isinstance(value, BaseFOV):
+            raise TypeError(
+                f"{self.__class__.__name__} value must subclass BaseFOV. "
+                f"Found {key} with value type {type(value)}"
+            )
+
+        if key in self:
```
ziw-liu, line 184: Is this redundant if `__init__` already receives a `dict`?

iohub/dataset_base.py
JoOkuma: @ziw-liu , I see. I was expecting `Position.scale` to return the same, but it might not be 5-dimensional, right? This might lead to confusion in the future.

iohub/fov.py
JoOkuma: It can be initialized in two different ways, so this additional check is necessary, and we only pay a very small overhead.

- kwargs init: `FOVDict(green=..., red=...)`
- dict init: `FOVDict({"green": ..., "red": ...})`
- both: `FOVDict({"green": ...}, red=...)`

iohub/fov.py
```diff
@@ -0,0 +1,222 @@
...
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> bool:
+        """Close the files."""
+        return True
```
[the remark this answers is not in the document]
JoOkuma: It should return `False`; thanks for noticing that.
[reference](https://stackoverflow.com/a/71171297/6748803)

iohub/fov.py
```diff
@@ -0,0 +1,222 @@
...
+        else:
+            is_tuple = False
+
+        for i in self._missing_axes():
+            seq.insert(i, value)
+
+        if is_tuple:
+            seq = tuple(seq)
+
+        assert len(seq) == len(_AXES_PREFIX)
```
ziw-liu, line 74:
```suggestion
        if len(seq) != len(_AXES_PREFIX):
            raise RuntimeError(f"Failed to pad raw axes {self.axes_names} to {_AXES_PREFIX}")
```

JoOkuma:
Thanks y'all.

PR #133, opened 2023-05-19
Convert a list of input datasets

talonchandler:
The current CLI for `convert` is `iohub convert -i <input-ome-tiff-folder> -o <output.zarr>`. 

In my experience/opinion this is a bit clunky because it requires you to
- type out the dataset name twice (once for input, once for output), 
- specify `.zarr` unnecessarily, and 
- it's not compatible with wildcards. 

This PR changes the interface to `iohub convert /glob/of/ome-tiff/folders/*` with an optional `-o <output-folder>` parameter that defaults to the current directory. 

In practice, I almost always want to convert all of the datasets within a folder, and I want to duplicate the folder names. So, for example, `iohub convert /path/to/project/*` will fill my current directory with the converted `.zarrs`.

### DIFF
--- a/iohub/cli/cli.py
+++ b/iohub/cli/cli.py
@@ -1,4 +1,5 @@
 import click
+import os
 
 from iohub._version import __version__
 from iohub.convert import TIFFConverter
@@ -19,7 +20,7 @@ def cli():
 @cli.command()
 @click.help_option("-h", "--help")
 @click.argument(
-    "files",
+    "datasets",
     nargs=-1,
     required=True,
     type=_DATASET_PATH,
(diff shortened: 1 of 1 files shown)


talonchandler:
Whoops now the real tests are breaking...I'll fix before re-requesting review.

ziw-liu:
How is this different from the following?

```sh
for dataset in /path/to/many/datasets/*; do iohub convert $dataset; done
```

talonchandler:
Thanks @ziw-liu, I agree. Since I wrote this I've improved my bash skills, and I'm fine leaving it as is. Closing.

PR #134, opened 2023-05-26
Emoji favicon

ziw-liu:
Use an emoji 💾 as the favicon. Other suggestions (or even better, a proper logo contribution) are very welcome.

Also attempting to fix the multiversion issue for the third time... This will still appear to be broken when building locally since the version switcher points to a static url that does not exist before this is merged.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -24,6 +24,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with: 
+          fetch-depth: 0
       - name: Setup Pages
         id: pages
         uses: actions/configure-pages@v3
@@ -34,7 +36,6 @@ jobs:
       - name: Build the docs
         working-directory: ./docs
         run: |
-          git fetch --all
           make publish
           touch build/html/.nojekyll
       - name: Upload build artifacts
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -34,6 +34,7 @@ jobs:
       - name: Build the docs
         working-directory: ./docs
         run: |
+          git fetch --all
           make publish
           touch build/html/.nojekyll
       - name: Upload build artifacts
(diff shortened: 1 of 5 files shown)


.github/workflows/docs.yml
```diff
@@ -18,10 +18,14 @@ concurrency:
 
 jobs:
   build:
+    env:
+      GITHUB_PAGES_URL: https://czbiohub.github.io/iohub
```
keithchev, line 22: this is not a huge deal but it would be more maintainable (particularly in light of our looming org rename) to construct this URL from the `GITHUB_REPOSITORY` env variable which afaik is defined by default in github actions. something like this:

```bash
REPO_OWNER="${GITHUB_REPOSITORY%%/*}";
REPO_NAME="${GITHUB_REPOSITORY#*/}";
GITHUB_PAGES_URL="https://$REPO_OWNER.github.io/$REPO_NAME";
```

docs/source/_static/switcher.json
```diff
@@ -1,10 +1,6 @@
 [
     {
         "version": "latest",
-        "url": "https://czbiohub/main/index.html"
-    },
-    {
-        "version": "0.1.0dev3",
-        "url": "https://mysite.org/en/v0.1.0dev3/index.html"
+        "url": "https://czbiohub.github.io/iohub/main/index.html"
```
keithchev, line 4: again it would be nice to avoid the hard-coded absolute URL here, but I can imagine that that might require dynamically generating this json file in `conf.py` or else injecting the URL into this file in the github action, and I can see that that might be too complicated/opaque

ziw-liu: I *think* I can get away with an absolute path without the domain name.

.github/workflows/docs.yml
ziw-liu: See https://github.com/czbiohub/iohub/pull/134/commits/1f3de567856cc534101afd8b2d7f581b03f6d082

keithchev: Nice!

keithchev (approved):
LGTM!

PR #137, opened 2023-05-31
Print number of positions in `iohub info`

talonchandler:
When I saw an `iohub info` result like this:
```
=== Summary ===
Format:		 omezarr v0.4
Axes:		 T (time); C (channel); Z (space); Y (space); X (space); 
Channel names:	 ['GFP EX488 EM525-45', 'mCherry EX561 EM600-37']
Row names:	 ['0']
Column names:	 ['0']
Wells:		 1
```
I didn't realize that the zarr store actually had multiple positions. 

This PR adds one line to the output to show the user the total number of positions:
```
Positions:	 5
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
 # basic pre-commit
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.1.0
+    rev: v4.4.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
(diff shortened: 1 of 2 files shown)


ziw-liu:
I didn't add this because every other field was in the plate metadata and is very quick to parse. Getting the total number of positions requires O(N) time and can be not very responsive if the dataset is large and/or the storage system has high latency.

Maybe move this to the verbose mode?

talonchandler:
Makes sense, I moved the message to `verbose` mode.

ziw-liu (approved):
Code looks good!

@JoOkuma can you check the pre-commit diff?

JoOkuma:
@ziw-liu LGTM, the update packages works for me too.

JoOkuma (approved):

PR #138, opened 2023-05-31
Remove state in NGFF plate initialization

### LINKED ISSUE #135: `create_position` fails after closing and reopening a dataset
This snippet demonstrates the issue:
```
from iohub import open_ome_zarr
dataset = open_ome_zarr('./test.zarr', layout="hcs", mode="a", channel_names=["1", "2"])
dataset.create_position("0", "0", "0") # WORKS
dataset.create_position("0", "1", "0") # WORKS
dataset.close()

# Reopen
dataset = open_ome_zarr('./test.zarr', layout="hcs", mode="a", channel_names=["1", "2"])
dataset.create_position("0", "2", "0") # FAILS
# ContainsGroupError: path '0' contains a group
```

This functionality is essential for slurm processing of zarr datasets, where a single node works on a single position and appends to an output zarr store when it finishes. Related: https://github.com/czbiohub/mantis/pull/33

ziw-liu:
Fix #135.

### DIFF
--- a/iohub/ngff.py
+++ b/iohub/ngff.py
@@ -1262,8 +1262,6 @@ class Plate(NGFFNode):
         self._acquisitions = (
             [AcquisitionMeta(id=0)] if not acquisitions else acquisitions
         )
-        self._rows = {}
-        self._cols = {}
 
     def _parse_meta(self):
         if plate_meta := self.zattrs.get("plate"):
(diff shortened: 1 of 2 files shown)


ziw-liu:
@talonchandler The actual problem is this:

https://github.com/czbiohub/iohub/blob/8418014ad83b0189e945e12cb0e2538d153c87e8/iohub/ngff.py#L1265-L1266

So even if it's being initialized for an existing store, the `Plate` object has this hidden internal state assuming that it's empty. So when a new well in an existing row is being created, it will try to create the existing row group again which causes the `ContainsGroupError`.

As for the `in` keyword, `zarr.Group.__contains__` works for any level of nesting. So row, well, and FOV paths will all work for a plate Zarr group.

talonchandler (approved):
Thanks for the explanation @ziw-liu. LGTM!

Issue #141, opened 2023-06-02
Tools for parallel `slurm` processing across positions


talonchandler:
The mantis project is depending on `iohub` in part because of the `zarr` format's ability to parallelize on SLURM. This issue is a request to clarify several behaviors of `iohub`, and to document best practices for parallel processing of hcs stores. 

**Question 1: When is plate-level and position-level metadata written?**

When multiple jobs try to write metadata to a .zarr store at the same time, race conditions can occur. @edyoshikun and I have been struggling to find a way to use `iohub` that avoids these race conditions. 

- Can we create an empty hcs zarr in advance, close the store, then write to it later? Has this been tested?
- When are you allowed to write to an existing .zarr and when are you not allowed? Any difference between hcs- and single-position .zarrs? How can we avoid `zarr.errors.ContainsGroupError: path '0/27/0' contains a group`-type errors?
- Why are `channel_names` supplied to `open_ome_zarr`? For example, is the following expected behavior?
```
python
Python 3.9.16 | packaged by conda-forge | (main, Feb  1 2023, 21:39:03) 
[GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from iohub import open_ome_zarr
>>> dataset = open_ome_zarr('./test.zarr', layout="hcs", mode="a", channel_names=["1", "2"])
>>> dataset.create_position("0","0","0")
<iohub.ngff.Position object at 0x7f70d4a61850>
>>> dataset.close()
>>> quit()
(mantis-dev) [talon.chandler@gpu-sm01-10] ~/iohub
19:20:42 $ iohub info test.zarr/
Reading file:	 /home/talon.chandler/iohub/test.zarr
WARNING:root:Zarr group at 0/0/0 does not have valid metadata for <class 'iohub.ngff.Position'>
WARNING:root:Cannot determine channel names: Invalid metadata at the first position
WARNING:root:Zarr group at 0/0/0 does not have valid metadata for <class 'iohub.ngff.Position'>
Traceback (most recent call last):
  File "/home/talon.chandler/.conda/envs/mantis-dev/bin/iohub", line 8, in <module>
    sys.exit(cli())
  File "/home/talon.chandler/.conda/envs/mantis-dev/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
[...]

PR #143, opened 2023-06-06
Gather FOVs into plates

ziw-liu:
~Proposed solution for #141 in https://github.com/czbiohub/iohub/issues/141#issuecomment-1575062339.~
Edit: The mantis project has decided to create the Zarr hierarchy before distributed processing. So this PR no longer requires review from that perspective. However, I think it is still useful for combining existing datasets in general, and this pooling can potentially benefit projects such as machine learning.

Usage pattern:

```py
from iohub.ngff import open_ome_zarr, Plate

# all FOVs from an HCS store
with open_ome_zarr("hcs.zarr") as old_plate:
    fovs = dict(old_plate.positions())

# FOV from a standalone FOV store
with open_ome_zarr("fov.zarr") as old_position:
    fovs["Z/1/0"] = old_position

new_plate = Plate.from_positions("combined.zarr", fovs)
```

### DIFF
--- a/iohub/ngff.py
+++ b/iohub/ngff.py
@@ -1239,6 +1239,79 @@ class Row(NGFFNode):
 class Plate(NGFFNode):
     _MEMBER_TYPE = Row
 
+    @classmethod
+    def from_positions(
+        cls,
+        store_path: StrOrBytesPath,
+        positions: dict[str, Position],
+    ) -> Plate:
+        """Create a new HCS store from existing OME-Zarr stores
+        by copying images and metadata from a dictionary of positions.
+
+        .. warning: This assumes same channel names and axes across the FOVs
+            and does not check for consistent shape and chunk size.
+
+        Parameters
+        ----------
+        store_path : StrOrBytesPath
+            Path of the new store
+        positions : dict[str, Position]
+            Dictionary where keys are destination path names ('row/column/fov')
+            and values are :py:class:`iohub.ngff.Position` objects.
+
+        Returns
+        -------
+        Plate
+            New plate with copied data
+
+        Examples
+        --------
+        Combine an HCS-layout store and an FOV-layout store:
+
+        >>> from iohub.ngff import open_ome_zarr, Plate
+
+        >>> with open_ome_zarr("hcs.zarr") as old_plate:
+        >>>     fovs = dict(old_plate.positions())
+
+        >>> with open_ome_zarr("fov.zarr") as old_position:
+        >>>     fovs["Z/1/0"] = old_position
+
+        >>> new_plate = Plate.from_positions("combined.zarr", fovs)
+        """
+        # get metadata from an arbitraty FOV
+        # deterministic because dicts are ordered
+        example_position = next(iter(positions.values()))
+        plate = open_ome_zarr(
+            store_path,
+            layout="hcs",
+            mode="w-",
+            channel_names=example_position.channel_names,
+            axes=example_position.axes,
+            version=example_position.version,
+        )
+        for name, src_pos in positions.items():
+            if not isinstance(src_pos, Position):
+                raise TypeError(
+                    f"Expected item type {type(Position)}, "
+                    f"got {type(src_pos)}"
+                )
+            name = normalize_storage_path(name)
+            if name in plate.zgroup:
+                raise FileExistsError(
+                    f"Duplicate name '{name}' after path normalization."
+                )
+            row, col, fov = name.split("/")
+            _ = plate.create_position(row, col, fov)
+            # overwrite position group
+            _ = zarr.copy_store(
+                src_pos.zgroup.store,
+                plate.zgroup.store,
+                source_path=src_pos.zgroup.name,
+                dest_path=name,
+                if_exists="replace",
+            )
+        return plate
+
     def __init__(
         self,
         group: zarr.Group,
--- a/tests/ngff/test_ngff.py
+++ b/tests/ngff/test_ngff.py
@@ -22,6 +22,7 @@ if TYPE_CHECKING:
 
 from iohub.ngff import (
     TO_DICT_SETTINGS,
+    Plate,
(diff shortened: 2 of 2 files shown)


mattersoflight:
An elegant API to create a blank OME HCS store with relevant metadata. Some questions:
@ziw-liu I'd suggest that you flesh out a solution for this [use case](https://github.com/czbiohub/iohub/issues/141#issuecomment-1574590415) before handing off the PR for testing.

edyoshikun:
Thanks @ziw-liu. Would we encounter this issue #64 if we copy a position that is too large?

ziw-liu:
> Thanks @ziw-liu. Would we encounter this issue #64 if we copy a position that is too large?

This method uses `zarr.copy_store()`, which copies the *files*, and does not deserialize the arrays. So as long as your file system is able to handle the chunk size it should work (don't use 4GB+ chunks on a 32 bit windows machine for example).

ziw-liu:
> An elegant API to create a blank OME HCS store with relevant metadata. Some questions: @ziw-liu I'd suggest that you flesh out a solution for this [use case](https://github.com/czbiohub/iohub/issues/141#issuecomment-1574590415) before handing off the PR for testing.

I'm suggesting a different approach to MP here. The example @talonchandler requested is creating a hierarchy before hand (step 2a, which is intentionally broken after creation) and populating the content afterwards (step 2b). This is against iohub's design where we try to make user-facing files verified against the specification as a quality assurance.

Using the utility provided in this PR, one would create any valid HCS *and/or* FOV layout OME-Zarr stores in parallel and would be able to combine arbitrary subsets of them into a new HCS store. In the SLURM scenario, one can process each position on a different node (with local python parallelism within the FOV), and gather the result as a last step. This is the typical scatter-and-gather scheme for HPC. Also in this process the datasets are valid OME-Zarr stores at every step, meaning that examination of intermediate result is possible.

Also, this gathering implementation allows for cherry-picking the FOVs, which is useful for partial analysis @ieivanov requested and also allows for pooling existing analysis for tasks like machine learning.

ziw-liu:
> 1. Pre-generate a very simple `ones.zarr`.

In my proposal, this step is not needed. The final HCS store does not need a template, but generated dynamically from the supplied path names. This is especially beneficial for general dataset pooling when processing is decoupled from gathering and multiple combinations are desired.

mattersoflight:
@ziw-liu you should prototype the solution that you think is widely useable across reconstruction and DL pipelines, but I think it is optimal if you write a small test to see how metadata is handled.

One idea in this direction: extend `test_combine_fovs_to_hcs` to `test_combine_fovs_to_hcs_mp` where a few processes read FOVs in parallel, do some compute, and write them to the shared hcs store. This will surface any design tradeoffs and bugs that we should be aware of first.

ziw-liu:
> What happens when you gather two fovs with conflicting metadata?

It is stated in the docstring as a warning that this is not checked. In many cases metadata 'conflicting' for specific tools like napari-ome-zarr are still valid according to the specification and downstream analysis, so there is no clear definition for it. In theory any position you are able to load in iohub (and passed to this function) is valid and does not conflict according to the specification when you combine them in HCS.

> What if they have the same name and/or well name?

You will not be able to construct the argument for the method because a dictionary cannot have two items with the same key.

@talonchandler In short the way parallelism is implemented does not change how this method works. It operates in a single thread on existing valid OME-Zarr datasets and constructs an arbitrary new collection of positions from them. When used in SLURM batch jobs, the downstream application will

edyoshikun:
@talonchandler @ziw-liu Will something like this work for our issue?

``` python
# scatter the positions
# based on our convention  i.e (input_pos_zarr = ./input.zarr/0/0/0)
def deskew(input_pos_zarr, output_pos_zarrr):
       with open_ome_zarr(input_pos_zarr') as position:
          # call the function for processing one position and return that
           deskewed_data = deskew(position)

        #Assuming the arrays are small, there is no need for a zarr store with zero arrays beforehand
        # If we need to create zero arrays make empty array per position with the appropriate shape
        tmp_out =./tmp/ + output_pos
        with open_ome_zarr(tmp_out, layout='fov') as fov_out:
           #you specify the data dimensions here
            image_array = fov_out.create_image('0', deskewed_data)

#Gather list of output positions after they are done...
#Create the dictionary of positions
fovs = {}
for p in list_output_positions:
    with open_ome_zarr(p) as position:
        fovs[p] = position

# Copy the data from ./tmp/output in the node to a "./combined/zarr" in luster
#the ./tmp can be discarted afterwards
new_plate = Plate.from_positions("combined.zarr", fovs)
```
I got a bit confused as to why can we not just generate the empty zarr with the proper shape beforehand? I imagine for the mantis skewed dataset `from mantis.analysis.deskew import get_deskewed_data_shape` does return the information to create a zero's array(?) I ran this here `/hpc/projects/comp.micro/mantis/2023_05_10_PCNA_RAC1/5-deskew_slurm_v2` and did write the metadata without a problem using the mentioned mantis function. 

@ziw-liu you Do you think this would work? If you need a dataset to test the issues we are having, here is a sample dataset: `/hpc/projects/comp.micro/mantis/2023_05_10_PCNA_RAC1/0-crop-convert-zarr/sample.zarr` which has a cropped version of `2023_05_10_PCNA_RAC` and code to generate whatever crop you want.

mattersoflight:
@ziw-liu , @JoOkuma This issue is a blocker for important analysis work that needs to happen for mantis project. And, I think the discussion here has surfaced two important needs and opportunities. So, let's iterate towards a slurm-compatible io. 
* In this iteration (next two days), please solve the problem of keeping metadata accurate that @edyoshikun , @talonchandler have surfaced. 
* In the next iteration (maybe after #132) is merged, let's make the pooling of FOVs available to multiple input formats. 

Thanks 🙏🏼

JoOkuma (approved):
LGTM!

PR #145, opened 2023-06-08
Update NDTIFF reader

### LINKED ISSUE #124: bump ndtiff version to >1.12.1
`ndtiff==1.12.0` and below have a bug which returned the wrong shape of dask arrays, see https://github.com/micro-manager/NDTiffStorage/pull/107

ziw-liu:
Resolves #124.

This bumps `ndtiff` to 2.1.0 so that we can use the automatic axes sorting instead of pulling our own.

Currently CI only tests for NDTIFFv2 datasets. @ieivanov Should we add some v3 datasets to the pool?

### DIFF
--- a/iohub/ndtiff.py
+++ b/iohub/ndtiff.py
@@ -140,24 +140,13 @@ class NDTiffReader(ReaderBase):
         # TODO: try casting the dask array into a zarr array
         # using `dask.array.to_zarr()`.
         # Currently this call brings the data into memory
-
-        ax = [
-            ax_
-            for ax_ in ["position", "time", "channel", "z"]
-            if ax_ in self._axes
-        ]
-
-        if "position" in self._axes.keys():
-            # da is Dask array
-            da = self.dataset.as_array(axes=ax, position=position)
-        else:
-            if position not in (0, None):
-                warnings.warn(
-                    f"Position index {position} is not part of this dataset."
-                    f" Returning data at default position."
-                )
-            da = self.dataset.as_array(axes=ax)
-
+        if "position" not in self._axes.keys() and position not in (0, None):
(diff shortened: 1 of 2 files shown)


iohub/ndtiff.py
```diff
@@ -140,32 +140,21 @@ def get_zarr(self, position: int) -> zarr.array:
         # TODO: try casting the dask array into a zarr array
         # using `dask.array.to_zarr()`.
         # Currently this call brings the data into memory
```
ieivanov, line 142: I believe now `as_array` returns a dask array without bring the data into memory, I think you can test and remove this comment

ieivanov (approved):
Haven't tested, but I think the code looks good

ieivanov:
> Should we add some v3 datasets to the pool?

Yes, I can acquire such datasets. Where would we store them? Should we update the waveorder collection or create a new one?

ziw-liu:
> Should we update the waveorder collection or create a new one?

Could work either way. If we do create a new one we could also migrate the existing test datasets over too.

iohub/ndtiff.py
ziw-liu: I *think* this means that `dask.array.to_zarr()` brings data to memory if the store is a `MemoryStore` so this method cannot be consistent with other `get_zarr()` methods.

ziw-liu: Hopefully this won't be a problem anymore after #132 and we refactor this.

ziw-liu:
@ieivanov do you think we should merge this as-is or wait for new test data?

ziw-liu:
Opened #149 to track the test dataset issue. Merging.

PR #146, opened 2023-06-12
Update URLs after organization migration

ziw-liu:
Fix links.


ziw-liu:
Also updated the link to documentation in repo short description to use GH pages automatically.

Issue #26, opened 2023-01-18
tensorstore support


JoOkuma:
Tensorstore could be used for the zarr backend.

A few advantages are:
- It's faster.
- It supports transposes and other transformations on the coordinate space without loading the data.
- It loads the data lazily until requested through `np.asarray`

PR #150, opened 2023-06-20
Add conversion from Zarr array to tensorstore

ilan-francisco:
Enabling tensorstore conversion from a zarr array, as mentioned in #26.

### DIFF
--- a/iohub/ngff.py
+++ b/iohub/ngff.py
@@ -5,6 +5,7 @@ import logging
 import math
 import os
 from copy import deepcopy
+from pathlib import Path
 from typing import TYPE_CHECKING, Generator, Literal, Sequence, Union
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


JoOkuma:
Thanks for the PR, @ilan-francisco.

I have a few minor comments:
- Could you add a test and document the method?
- You also need to add the package as a dependency. If it's an optional dependency, `pytest` can skip the test if the import fails.
- Convert the `assert` to an `if` and `raise` call.

@ziw-liu , will `tensorstore` be an optional dependency?

ziw-liu:
> will `tensorstore` be an optional dependency?

This sounds like the best solution as the complied extension of tensorstore is not available on every platform.

If we do go this route then this PR (and future ones using ts) will need to handle import errors.

iohub/ngff.py
```diff
@@ -340,7 +341,19 @@ def downscale(self):
         raise NotImplementedError
 
     def tensorstore(self):
-        raise NotImplementedError
+        import tensorstore as ts
+
+        assert self.path != "", "Zarr array must be stored on a file system."
```
ziw-liu, line 346: In theory an iohub NGFF store is always on a local FS, so this might be redundant.

ziw-liu:
@JoOkuma Is further follow-up work expected for this PR?

ziw-liu:
Closing as stale.

PR #154, opened 2023-06-22
Improve CC gallery example

JoOkuma:


ziw-liu:
You would need to change the file name to `run_*.py` if this needs to be executed.

(Not that it's going to work since we don't have napari installed.)

ziw-liu:
Maybe do this?

```py
try:
    import napari
except ImportError:
    sys.exit(0)
viewer = ...
napari.run()
```

docs/source/conf.py
```diff
@@ -209,7 +210,7 @@
     "examples_dirs": "../examples",
     "gallery_dirs": "auto_examples",
     "download_all_examples": False,
-    "filename_pattern": "/run_",
+    # "filename_pattern": "/run_",
```
ziw-liu, line 213: The default is `'/plot'`, not everything.

JoOkuma:
@ziw-liu 
> (Not that it's going to work since we don't have napari installed.)
It's working with `sphinx-gallery` in an environment without napari, it just can't be called as a `__main__` file.

ziw-liu:
> @ziw-liu
> 
> > (Not that it's going to work since we don't have napari installed.)
> 
> It's working with `sphinx-gallery` in an environment without napari, it just can't be called as a `__main__` file.

Rendering will always work (and worked before this PR). My point was that the blocks before importing napari could still be executed (like the first blocks in the large array example) so that the output of loading times (L43 and L50) can be displayed.

docs/source/conf.py
ziw-liu: Will need to undo this change before merging.

ziw-liu (approved):
This won't be run in CI for now. If the filename is changed to `run_view_clearcontrol_dataset.py` it will run and the last block should error like [this](https://sphinx-gallery.github.io/stable/auto_examples/no_output/plot_raise.html), which is probably fine.

Issue #142, opened 2023-06-05
Place FOVs in correct HSC row/col upon zarr conversion based on metadata


ieivanov:
The mantis acquisition engine generates [metadata](https://github.com/czbiohub/mantis/blob/main/docs/data_structure.md) on which position comes from which well in a 96 well format, currently using a [positions.csv](https://github.com/czbiohub/mantis/blob/main/docs/positions.csv) file. It will be helpful if `iohub convert` can take this file as input and write HCS data in the appropriate rows/columns. I think we can modify the `grid_layout` argument to `TIFFConverter` to be `bool` or `dict` in `{position_index: [row_label, col_label]}` format, e.g. `{0: [A, 1], 1: [A, 1], 2: [A, 2]}`, etc. 

Related issue: https://github.com/micro-manager/pycro-manager/issues/575. The use of a `positions.csv` file is temporary until we can encode the FOV label in the pycromanager metadata. Maybe for now `iohub convert` can take the `positions.csv` file as an input until we can read this information from the metadata.

ziw-liu:
The `positions.csv` file doesn't seem to exist in the data structure example directory of mantis or recent acquisitions. @ieivanov can you provide an example where it's used?

ieivanov:
An example `positions.csv` file is provided here: https://github.com/czbiohub-sf/mantis/blob/main/docs/positions.csv. An example dataset using the `positions.csv` file is: `comp_micro/rawdata/mantis/2023_04_05_mantis_HEK/48wells_1timepoint_4`

PR #159, opened 2023-07-12
Convert plate layout from Micro-Manager OME-TIFFs

ziw-liu:
When the HCS position generator plugin in MM is used, there is enough information in the stage position labels (e.g. `'A1-Site_0'`) to parse the intended well layout for NGFF HCS.

### Interface change

- `MicromanagerOmeTiffReader` now has a `hcs_position_labels` property for parsing this metadata
- `TIFFConverter` can now be initialized with `hcs_plate=True`
- The `iohub convert` CLI command now has a `--hcs-plate`/`-t` flag

### New problems

- `-t` is arguably a bad shortcut, suggestions welcome!
- This is not tested automatically, since we don't have small test files that have this metadata
- The pattern is hard-coded, and may not account for variations from different MM/plugin versions

Related: #142


ziw-liu:
Hitting pydantic 2.0 issues here...

talonchandler (commented):
This is looking great! Thanks so much @ziw-liu. 

I've tested on two datasets from hummingbird (`/hpc/instruments/cm.hummingbird/Grace/20230615_hek293T_pcna_ro3306_3` and `/hpc/instruments/cm.hummingbird/Michelle/2023_06_21_ZikaInfected_OrganelleRemodeling/pos_1 `) and it's met all of my expectations so far. I'm letting the conversions run, and I'll double check that everything is in the correct place when these conversions finish. 

My only questions so far are minor:

- For the second dataset `iohub info` gives:
```
Channel names:	 ['BF', 'DAPI', 'FITC']
Row names:	 ['B']
Column names:	 ['3', '2', '1']
```
Can you confirm that "Channel names" are ordered (corresponding to the `C` dimension in the 5D array), while the "Column names" aren't strictly ordered for the output zarr store (although the order does correspond to the acquisition order)? I'm not sure if I'm suggesting a change here...just double checking. Maybe `{` instead of `[` to distinguish unordered vs. ordered in `iohub info`?

- I'm also seeing a non-natsorted tree like this:
```
/
 └── B
     ├── 1
     │   ├── 0
     │   │   └── 0 (17, 3, 31, 2048, 2048) uint16
     │   ├── 1
     │   │   └── 0 (17, 3, 31, 2048, 2048) uint16
     │   ├── 10
     │   │   └── 0 (17, 3, 31, 2048, 2048) uint16
     │   ├── 11
     │   │   └── 0 (17, 3, 31, 2048, 2048) uint16
     │   ├── 12
     │   │   └── 0 (17, 3, 31, 2048, 2048) uint16
```
which I'm guessing comes from zarr's `tree` function. Any simple way to natsort this output? Not a high priority if it's tricky. 

---
Regarding the naming, can you improve the help strings for these two options? I haven't played with `-g`, so it's not obvious to me how they're different. 

```
-g, --grid-layout           Arrange positions in a HCS grid layout
-t, --hcs-plate             Arrage FOVs according to the MM HCS
                              plate position labels
```
---
I would vote for the behavior you're testing here to be on by default (at least for OME-Tiff datasets).

ziw-liu:
> Can you confirm that "Channel names" are ordered (corresponding to the `C` dimension in the 5D array), while the "Column names" aren't strictly ordered for the output zarr store (although the order does correspond to the acquisition order)?

Yes this is expected. I don't know if we should sort this, since acquisition order is the only 'intended' order we can get.

ziw-liu:
> Any simple way to natsort this output?

Not that I know of. The zarr function prints to stdout directly. Also not really a big concern IMO because that's the same output from `ls` or `tree`.

ziw-liu:
> I would vote for the behavior you're testing here to be on by default (at least for OME-Tiff datasets).

The metadata used here is not always available. By 'on by default' are you suggesting a silent fallback when pattern matching fails?

ziw-liu:
> You might even consider getting rid of the `-t/--hcs-plate` flag...can you think of a case where you'd want to turn this off?

Good point. The only hypothetical scenario I can think of is when the labels match the pattern (`-Site_`) but are not actually valid for HCS (having typos or generated by other plugins). But removing the CLI flag is still OK because this can be rescued by using the Python API directly.

ziw-liu:
> I'm happy to approve now or later. Let me know what your plans are with this PR.

I'm trying to come up with some CI tests for the new code. Will probably need to hack it a little bit since it's not in the test datasets, unless there's a downloadable small example somewhere on the internet.

ieivanov:
I'm looking at this now, I can make you a small test dataset

ieivanov (approved):

Issue #164, opened 2023-07-24
Single-page TIFF converter regression?


ziw-liu:
CI runs after merging #159 has been [failing](https://github.com/czbiohub-sf/iohub/actions/runs/5648283652) for the single-page TIFF converter. However I was not able to reproduce locally. Also that PR shouldn't have changed how single-page TIFFs work (tests passed before merging).

ziw-liu:
Seems to be a zarr 2.15->2.16 problem.

ziw-liu:
#165 pins zarr version, but this should be reverted after a proper fix is in place.

PR #165, opened 2023-07-26
Pin zarr version to avoid tifffile issue

ziw-liu:
Zarr 2.16 introduces a change related to tifffile. Since we pinned tifffile to an older version this may have broke our TIFF converters.

Temporarily mitigates #164


ziw-liu:
Let's keep #164 open util a proper fix (updating TIFF readers to use the new tifffile).

talonchandler (approved):

PR #170, opened 2023-08-08
Convert image metadata for ND-TIFF

### LINKED ISSUE #168: Convert image plane metadata
@ieivanov raised the need of converting all image plane metadata, which will be useful for the mantis project.

In a previous meeting we agreed that the easiest, fastest, and most robust way to preserve MM image plane metadata during conversion is to toggle on the option to store them as a separate TXT file in MM. However this may not be accessible from pycromanager. @ieivanov is this the reason why need such a feature? If so I propose that we only implement this for ND-TIFF, and write an output TXT file similar to what MM MDA would produce.

ziw-liu:
Converts ND-TIFF image plane metadata to `converted.zarr/ndtiff_metadata.json`.

Resolves #168.

### DIFF
--- a/docs/source/api/mm_converter.rst
+++ b/docs/source/api/mm_converter.rst
@@ -1,13 +1,13 @@
 Convert TIFF to OME-Zarr
 ========================
 
-.. currentmodule:: iohub.convert
-
 
 .. note:: There is also a CLI command for conversion.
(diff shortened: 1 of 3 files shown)


ieivanov (approved):
LGTM. Made a couple of small changes

PR #171, opened 2023-08-14
Convert ND-TIFF position label

ziw-liu:
Convert non-HCS string-valued ND-TIFF position coordinate into column names of an HCS OME-Zarr store.

### DIFF
--- a/iohub/cli/cli.py
+++ b/iohub/cli/cli.py
@@ -82,14 +82,7 @@ def info(files, verbose):
     is_flag=True,
     help="Arrange FOVs in a row/column grid layout for tiled acquisition",
 )
-@click.option(
-    "--label-positions",
-    "-p",
-    required=False,
-    is_flag=True,
-    help="Dump postion labels in MM metadata to Omero metadata",
-)
-def convert(input, output, format, scale_voxels, grid_layout, label_positions):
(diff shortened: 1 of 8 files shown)


ziw-liu:
@ieivanov The test v3 dataset does not have channel names. Is this intentional or an NDTiffStorage bug?

ieivanov:
`ndtiff_v3_labeled_positions` only has a position dimension, no channels and therefore no channel names

ieivanov:
@ziw-liu can you please test on `cm.automaton/mantis_HCS_acq_1` and debug? I am getting a `ValueError: Image coordinate channel = 1 is not part of this dataset.` error. The string axis parsing needs to be extended to the channel dimension as well.

ziw-liu:
> @ziw-liu can you please test on `cm.automaton/mantis_HCS_acq_1` and debug? I am getting a `ValueError: Image coordinate channel = 1 is not part of this dataset.` error. The string axis parsing needs to be extended to the channel dimension as well.

@ieivanov I can now convert those datasets. Can you review again?

ieivanov (approved):
I think this is good for now. Thanks a lot!

When you're working on the readers next, you can consider moving `_normalize_ndtiff_coord` to the reader to simplify `_check_coordinates` and the metadata fetching during init which still uses integer indices.

PR #175, opened 2023-08-16
Only set array transformation upon creation of FOVs

ziw-liu:
Set FOV-level transformation to identity upon creation of the first array.


edyoshikun (approved):
LGTM! Thanks @ziw-liu

talonchandler (approved):

PR #178, opened 2023-08-22
Enable XYZ chunking at conversion

### LINKED ISSUE #176: `iohub convert` chunks by XY planes rather than XYZ volumes
Was this an intentional decision? I see the docstring says that using XY planes is fastest at conversion time, but that doesn't align with the rest of our processing pipelines which chunk by XYZ volumes.

### LINKED ISSUE #177: `iohub convert` would be more efficient if it reads whole XYZ volumes at a time
Conversion to zarr is turning our to be one of the slower steps in our analysis pipelines.

Currently `iohub convert` reads and writes one image at a time using a call to `reader.get_image(P, T, C, Z)`. The conversion would be faster if we read and write whole XYZ volumes at a time using `reader.get_zarr(P)[T, C]`.

ieivanov:
This PR enables XYZ chunking at zarr conversion and speeds up `ndtiff` conversion following [this](https://github.com/czbiohub-sf/compmicro-sandbox/blob/master/iohub_scripts/convert/convert_isim_fish.py) example.

Fixes #176, Fixes #177 

To use, call `iohub convert -i /path/to/input/ -o /path/to/output.zarr/ --chunks XYZ`. Z stacks may be split into multiple chunks if they are larger than `MAX_CHUNK_SIZE` bytes.

We should use this code to convert future datasets and we should also re-convert older large datasets to reduce the number of small files on our storage.

Let's build on this code and make the converter work well. So far, I've tested it only with `pytest` and latest mantis `ndtiff` datasets. Currently, the converter effectively splits into two paths, depending on whether you're converting a `ndtiff` dataset or now. I believe @ziw-liu's intension is to make all readers return a lazy dask array, similarly to how `ndtiff` works. This can be a starting point to a universal converter following the revamp of the readers as part of the Universal API.

TODO:
 - [x] Test that missing timepoints are handled correctly

### DIFF
--- a/iohub/cli/cli.py
+++ b/iohub/cli/cli.py
@@ -101,3 +101,6 @@ def convert(input, output, format, scale_voxels, grid_layout, chunks):
         chunks=chunks,
     )
     converter.run()
+
+if __name__ == '__main__':
+    convert()
--- a/iohub/cli/cli.py
+++ b/iohub/cli/cli.py
@@ -101,6 +101,3 @@ def convert(input, output, format, scale_voxels, grid_layout, chunks):
         chunks=chunks,
     )
     converter.run()
-
-if __name__ == '__main__':
-    convert()
--- a/iohub/cli/cli.py
+++ b/iohub/cli/cli.py
@@ -88,7 +88,7 @@ def info(files, verbose):
     required=False,
     default="XY",
     help="Zarr chunk size given as 'XY', 'XYZ', or a tuple of chunk "
-    "dimensions.",
+    "dimensions. If 'XYZ', chunk size will be limited to 500 MB.",
 )
 def convert(input, output, format, scale_voxels, grid_layout, chunks):
     """Converts Micro-Manager TIFF datasets to OME-Zarr"""
--- a/iohub/cli/cli.py
+++ b/iohub/cli/cli.py
@@ -87,7 +87,7 @@ def info(files, verbose):
     "-c",
     required=False,
     default="XY",
-    help="Zarr chunk size given as, 'XY', 'XYZ', or a tuple of chunk "
+    help="Zarr chunk size given as 'XY', 'XYZ', or a tuple of chunk "
     "dimensions.",
 )
 def convert(input, output, format, scale_voxels, grid_layout, chunks):
--- a/iohub/cli/cli.py
+++ b/iohub/cli/cli.py
@@ -82,7 +82,15 @@ def info(files, verbose):
     is_flag=True,
     help="Arrange FOVs in a row/column grid layout for tiled acquisition",
 )
-def convert(input, output, format, scale_voxels, grid_layout):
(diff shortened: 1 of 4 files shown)


iohub/convert.py
```diff
@@ -447,44 +472,85 @@ def run(self, check_image: bool = True):
...
-                p, t, c, z = self._normalize_ndtiff_coord(*coord_reorder)
-            else:
-                p, t, c, z = coord_reorder
-            img_raw = self._get_image_array(p, t, c, z)
-            if img_raw is None or not getattr(img_raw, "shape", ()):
-                # Leave incomplete datasets zero-filled
-                logging.warning(
-                    f"Cannot load image at PTCZ={(p, t, c, z)}, "
-                    "filling with zeros. Check if the raw data is incomplete."
+            for p_idx in tqdm(range(self.p), bar_format=bar_format):
```
ziw-liu, line 477: Can this separate path be contained in its own method?

iohub/convert.py
```diff
@@ -447,44 +472,85 @@ def run(self, check_image: bool = True):
...
+                try:
+                    ndtiff_pos_idx, *_ = self.reader._check_coordinates(
+                        ndtiff_pos_idx, 0, 0, 0
+                    )
+                except ValueError:
+                    # Log warning and continue if some positions were not
+                    # acquired in the dataset
+                    logging.warning(
+                        f"Cannot load data at position {ndtiff_pos_idx}, "
+                        "filling with zeros. Raw data may be is incomplete."
```
ziw-liu, line 492:
```suggestion
                        "filling with zeros. Raw data may be incomplete."
```

edyoshikun (approved):
LGTM!

iohub/cli/cli.py
```diff
@@ -82,13 +82,22 @@ def info(files, verbose):
...
     help="Arrange FOVs in a row/column grid layout for tiled acquisition",
 )
-def convert(input, output, format, scale_voxels, grid_layout):
+@click.option(
+    "--chunks",
+    "-c",
+    required=False,
+    default="XY",
+    help="Zarr chunk size given as 'XY', 'XYZ', or a tuple of chunk "
+    "dimensions.",
```
talonchandler, line 91: Am I reading correctly that 'XYZ' chunks will be clipped to 500 MB? If yes, I think this is a good choice and should be documented:

```suggestion
    "dimensions. If 'XYZ', chunk size will be limited to 500 MB.",
```

talonchandler (approved):

ziw-liu (changes requested):
Holding for https://github.com/czbiohub-sf/iohub/pull/178#discussion_r1300781300

ziw-liu (approved):

ieivanov:
Tested that missing timepoints are now handled correctly, ready to merge

PR #182, opened 2023-08-24
Change exception to warning if xy_scale_size metadata is absent

### LINKED ISSUE #180: `iohub info` fails on MM datasets without `xy_pixel_size`
...and `xy_pixel_size` is absent in our 1.3.0 zenodo datasets (and oftentimes from the microscopes if the pixel size isn't calibrated). 

For example:
```
$ iohub info 2022_08_04_recOrder_pytest_20x_04NA/2T_3P_16Z_128Y_256X_Kazansky_1/
Reading file:	 /Users/talon.chandler/Desktop/0.4.0-release/misc/recOrder_test_data/2022_08_04_recOrder_pytest_20x_04NA/2T_3P_16Z_128Y_256X_Kazansky_1
  File "/opt/anaconda3/envs/recorder-dev/bin/iohub", line 8, in <module>
    sys.exit(cli())
  File "/opt/anaconda3/envs/recorder-dev/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
  File "/opt/anaconda3/envs/recorder-dev/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/opt/anaconda3/envs/recorder-dev/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
  File "/opt/anaconda3/envs/recorder-dev/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
  File "/opt/anaconda3/envs/recorder-dev/lib/python3.10/site-packages/click/core.py", line 783, in invoke
  File "/Users/talon.chandler/iohub/iohub/cli/cli.py", line 43, in info
    print_info(file, verbose=verbose)
  File "/Users/talon.chandler/iohub/iohub/reader.py", line 267, in print_info
    reader.xy_pixel_size,
  File "/Users/talon.chandler/iohub/iohub/multipagetiff.py", line 309, in xy_pixel_size
    raise AttributeError("XY pixel size cannot be determined.")
AttributeError: XY pixel size cannot be determined.. Did you mean: '_xy_pixel_size'?
```

talonchandler:
Fixes #180. 

Also, defaults to 1.0 um if metadata is absent.

### DIFF
--- a/iohub/multipagetiff.py
+++ b/iohub/multipagetiff.py
@@ -306,7 +306,10 @@ class MicromanagerOmeTiffReader(ReaderBase):
     def xy_pixel_size(self):
         """XY pixel size of the camera in micrometers."""
         if self._xy_pixel_size is None:
-            raise AttributeError("XY pixel size cannot be determined.")
+            logging.warning(
+                "XY pixel size cannot be determined. Returning 1.0 um."
+            )
+            return 1.0
         return self._xy_pixel_size
 
     def _get_dimensions(self, position):


ziw-liu:
We can also modify the `info` command to catch the error and show 'not available'. The main concern I have is user code depending on the old behavior will silently use the wrong value if it's not executed in an interactive environment.

talonchandler:
Possibly naive, but is it possible for users to depend on the old behavior if it raises `AttributeError` when the metadata is absent?

I also think that setting to default 1.0 is the right thing to do here since that's what happens for the ome-zarr files with identity transformations, right? 

FWIW this code was introduced ~3 months ago in #118.

ziw-liu:
> is it possible for users to depend on the old behavior if it raises `AttributeError` when the metadata is absent?

Something like:

```py
try:
    px_size = reader.xy_scale_size
except AttributeError:
    px_size = json.load("some/custom/metadata.txt")["pixel_size"]
```

I don't know anyone who is doing this, but if the purpose is just to fix #180 there is also no need to break APIs.

ziw-liu:
Closing in favor of #185 and will be shipped in v0.2.0.

Thanks @talonchandler!

Issue #157, opened 2023-06-29
Root path of BaseFOV


ziw-liu:
In #132 the `BaseFOV.root` returns a `pathlib.Path` object and `__eq__` compares the paths of two FOVs to check if they are the same. This assumes that an FOV always has a single path. However in OME-TIFF and NDTIFF one FOV can be distributed in many TIFF files (usually a subset of all the TIFF files in the dataset).
Edit: the same file can also contain multiple FOVs.

@JoOkuma Should we change this or is there a way around it?

ziw-liu:
Also for OME-TIFF and NDTIFF it is cumbersome for the FOV objects to carry state of the index map, because parsing has to happen at the dataset (`BaseFOVMapping`) level.

ziw-liu:
@JoOkuma Do you have any thoughts on this? I would like to resolve it before releasing.

Issue #155, opened 2023-06-22
Per-frame metadata in UAPI


ziw-liu:
In https://github.com/czbiohub-sf/iohub/pull/132#discussion_r1224842821 (@talonchandler) and https://github.com/czbiohub-sf/iohub/issues/103#issuecomment-1504154309 (@JoOkuma) the need for accessing each time stamp in a time lapse image is surfaced. However, implementing a clean API at the FOV level is not very straightforward, so we settled on a scalar (uniform delta) for FOVs.

In https://github.com/czbiohub-sf/iohub/issues/151 a more general feature is requested by @ieivanov to access metadata specific to each frame.

IMO the XY frame is indeed a physically grounded entity (readout from the 2D camera sensor), and accessing metadata such as exact time stamp and hardware settings can be useful. It may reduce confusion if these is a shared API across file formats, but I think it adds too much complexity to let frames have their own type, so I'm imagining `BaseFOV` methods that will return these information given the coordinates. For example:

```py
BaseFOV.frame_camera_timestamp(T: int, C: int, Z: int) -> datetime.datetime:
    ...
```

And MMStack readers can have their own base class providing the specific hardware information.

ziw-liu:
One caveat is that these information is not always available. So they cannot be guaranteed as other #132 attributes are.

ieivanov:
The `ZarrReader` provides `get_image_plane_metadata` and the `NDTiffReader` provides `get_image_metadata` which return image plane metadata. They also have methods to return the micro-manager summary metadata, which is metadata for the entire acquisition (multiple FOVs). `ClearControlFOV` provides `metadata` which, as I read it, is also per acquisition.

It will be useful if the UAPI specifies how to handle per-image metadata (e.g. image timestamp, state of microscope hardware at this image), per-FOV metadata (e.g. FOV XYZ coordinates, FOV label), and per-dataset metadata (e.g. data dimensions, microscope name, user, sample conditions, etc.)

talonchandler:
I really like this signature you're proposing @ziw-liu:
```
BaseFOV.frame_camera_timestamp(T: int, C: int, Z: int) -> datetime.datetime:
```
Building on this, maybe the **per-frame metadata** can be accessible via:
```
BaseFOV.frame_metadata(T: int, C: int, Z: int) -> Dict[str, Any]
```
the **per-FOV metadata** via:
```
BaseFOV.metadata -> Dict[str, Any]
```
and the **per-dataset metadata** via:
```
[...]

PR #185, opened 2023-09-02
Universal API implementations for Micro-Manager OME-TIFF and NDTiff

### LINKED ISSUE #53: Improve TIFF I/O stack
For now we use a combination of `tifffile`, custom code (multi-page OME-TIFF), and `ndtiff` (pycromanager NDTIFF) to read TIFF-based formats. Given that `tifffile` also supports NDTIFF, there may be room to simplify the reading stack.

Also, we should investigate if `tifffile` and AICSImageIO has improved their OME-TIFF reading performance.

@nclack also mentioned that there are additional strategies we can explore to accelerate TIFF I/O.

### LINKED ISSUE #85: Refactor `_set_mm_meta` in `BaseReader` and its children
This discussion started on #71.

Multiple classes have extremely similar `_set_mm_meta` and other micromanager-related methods that could be simplified and reduce code duplication.

Additionally, why do non-micromanager classes need special micromanager attributes? Shouldn't they live with the rest of the metadata?

### LINKED ISSUE #151: The MicromanagerOmeTiffReader needs methods to retrieve the image plane metadata
The MicromanagerOmeTiffReader needs methods to retrieve the image plane metadata. The internal method `_infer_image_meta` can be modified/extended to retrieve all data in the `MicroManagerMetadata` tif tag as a dictionary given the image PTCZ coordinates.

### LINKED ISSUE #152: The MicromanagerOmeTiffReader needs better definitions or documentation for position attributes
The MicromanagerOmeTiffReader provides `positions`, `stage_positions`, `position_arrays`, and `get_num_positions` attributes / methods. We need better definition and/or documentation for these. Some (like `position_arrays`) should probably be internal and others (like `get_num_positions` which returns `self.positions`) should probably be deprecated.

### LINKED ISSUE #164: Single-page TIFF converter regression?

### LINKED ISSUE #180: `iohub info` fails on MM datasets without `xy_pixel_size`

### LINKED ISSUE #207: unpin zarr versions
Hi, is it time to revert the change done in #165?

How can I check if `zarr<2.16` is still a requirement?

ziw-liu:
Implements universal API (UAPI) for MMStack (Micro-Manager OME-TIFF) and NDTiff (written by MM and pycro-manager). This PR contains many breaking changes and is geared towards the v0.2.0 release.

### Unresolved issues

- UAPI for OME-NGFF
- #157 (@JoOkuma)

#### API

- UAPI entry point (#40): rename `read_micromanager` to `read_images`
  - Removed `extract_data` and `format` arguments
  - Suggestions to function name are welcome!
  - TODO: include clear control?
- UAPI for MM OME-TIFF
  - Replace `multipagetiff.MicromanagerOmeTiffReader` with `mmstack.MMStack`
    - [x] Closes #152 (@ieivanov)
    - [x] Closes #85 (@JoOkuma)
  - When the channel names from metadata is incomplete (e.g. multi-camera mode on the Dragonfly microscope), the name list will be padded with generic names, e.g. `[..., "Channel2", "Channel3", ...]`. Note that this will be *incorrect* if the partially available channel names do not start with the first channel or if they are not contiguous. A warning will be logged and the user is responsible to correct this downstream.
- UAPI for NDTiff
  - Replace `ndtiff.NDTiffReader` with `ndtiff.NDTiffDataset`
  - When channel names are not available from metadata, fallback to `['Channel0', 'Channel1', ...]`  instead of `['0', '1', ...]`
- OME-TIFF and NDTiff FOV objects now have a `frame_metadata` object as proposed in #155
  - [x] Closes #151 (@ieivanov)
- Non-UAPI readers (MM single-page tiff, v0.1 OME-Zarr, uPTI) are moved into the `_deprecated` namespace
  - Only accessible from specific API (importing corresponding reader classes)
  - Testing is still maintained and there is no immediate plan to remove them
- When ZYX scaling information is not available from OME-TIFF metadata, the reader will fall back to `1.0` and log a warning.
  - [x] Closes #180 (@talonchandler)
  - [x] Replaces #182 (@talonchandler)
- Renamed `convert.TIFFConverter.run()` to `convert.TIFFConverter.__call__()`
 
#### CLI

- Removed support for converting single-page TIFFs
- Removed arguments from the `convert` command
  - `--format` is always auto-inferred
  - `--scale-voxels` is always TRUE
  - `--check-image` is always FALSE
- The chunk size argument `--chunks` now works for both NDTiff and OME-TIFF
  - The default chunk size is ZYX for all input formats
- Avoid using the root logger and set the default logging level to INFO (#188)
  - The logging level can be changed with an environment variable, e.g. `IOHUB_LOG_LEVEL=DEBUG`

#### Documentation

- Updated sphinx theme
- TODO: add version switcher for old releases

### Technical changes

- Use tifffile's `TiffZarrStore` to acheive lazy loading of OME-TIFF. The custom index map parsing is removed.
  - [x] Closes #53
- Use xarray to provide named axes in the TCZYX order. (The `xdata` attribute gives access to the underlying xarray object.)
- Bumps tifffile and zarr-python to latest
  - [x] Closes #207 (@JoOkuma)
  - [x] Closes #164
  - [x] Replaces #208 (@ilan-francisco)
- Reader testing is now done for all the applicable files
  - Replaced session-scoped fixtures in `conftest.py` with globals since parametrization happens before fixture is available

### Notes

Many issues are linked to be closed by this PR.
Reviewers: please check the boxes and leave a sign-off when an issue has been verified to be fixed.


ziw-liu:
@JoOkuma Not sure if this is going to work for the DaXi data since I don't have 1023x2048x2048 volumes to test with...
Also if you don't like the Dask conversion, the `TiffZarrStore` (in acquisition axes ordering) is available through `MMStack._store`.

JoOkuma:
@ziw-liu, thank you. I'll give it a try soon

ziw-liu:
Fixing docs...

ziw-liu:
@talonchandler channel names are back! Chunk sizes now prints by default for single FOVs and is hidden behind `--verbose` for plates.

JoOkuma:
Hi @ziw-liu , thanks for moving forward with this.

I'll take a look at it soon.

ieivanov:
Very exciting indeed, I'll start testing too.

iohub/reader.py
```diff
@@ -65,75 +41,57 @@ def _check_zarr_data_type(src: str):
...
-    # pick parent directory in case a .tif file is selected
-    if src.endswith(".tif"):
-        src = os.path.dirname(src)
-
-    files = glob.glob(os.path.join(src, "*.tif"))
+def _check_single_page_tiff(src: Path):
+    if src.is_file():
+        src = src.parent
+    files = src.glob("*.tif")
     if len(files) == 0:
```
JoOkuma, line 48: Could you check if `len(files)` work?
I think it will return ```TypeError:` object of type 'generator' has no len()```

JoOkuma (commented):
Hi @ziw-liu , great PR! I left a few comments.

I notice you're using `array`. Should we adopt this to all other classes too?
I'm unfamiliar with it, but I'm happy to adopt it with our formats if it's zarr-like.

iohub/convert.py
```diff
@@ -72,27 +67,22 @@ def _create_grid_from_coordinates(
 
 class TIFFConverter:
```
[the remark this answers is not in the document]
ziw-liu: This currently relies on micro-manager metadata to determine plate layout and write acquisition metadata. But you are right that another FOV mapping can be made to work. We don't currently have any concrete example of that though.

iohub/reader.py
ziw-liu: Thanks for catching this! Also added a new test.

ziw-liu:
> I notice you're using array. Should we adopt this to all other classes too?

@JoOkuma I think the `ArrayLike` return type is OK. I'm using Xarray since Dask has to be used for TIFFs anyway. In other cases Dask may not be ideal for performance reasons.

ieivanov:
I really like this implementation! I'll post messages as I continue reading and testing and I'll provide an official review at the end.

> Suggestions to function name are welcome!

Renaming suggestion - currently we have the following class names

`BaseFOV` : `BaseFOVMapping`
`MicroManagerFOV` : `MicroManagerFOVMapping`
`MMOmeTiffFOV` : `MMStack`
`NDTiffFOV` : `NDTiffDataset`

it may be more "symmetric" to name the classes as follows:

`BaseFOV` : `BaseFOVStack`
`MicroManagerFOV` : `MicroManagerFOVStack`
`MMOmeTiffFOV` : `MMStack`
`NDTiffFOV` : `NDTiffStack`

`stack` could be replaced with `dataset`, if you prefer. One more iteration could be:

`BaseFOV` : `BaseFOVStack`
`MicroManagerFOV` : `MicroManagerFOVStack`
`OMETiffFOV` : `OMETiffStack`
`NDTiffFOV` : `NDTiffStack`

`OMETiffFOV` inherits from `BaseFOV` and `OMETiffStack` inherits from `MicroManagerFOVStack`. We likely won't be reading other types of ome.tif files.

ziw-liu:
@ieivanov with `MM` in the name of OME-TIFF reader classes I'm hoping to note that there is no support for a spec-adherent OME-TIFF dataset that is not the direct output of MM.

ieivanov (changes requested):
Again, I think this is a great improvement to `iohub`.

Before merging, we should update `readme.md` to reflect the new `read_images` API. It would also be very helpful to include examples of the new slicing and string-indexing that xarray enables. Most/all of the official `iohub` examples currently focus on the zarr reader/writer.

It would also be very helpful to include `__repr__` methods for `MicroManagerFOV`, `MicroManagerFOVMapping`, `NDTiffFOV`, and `NDTiffDataset`, largely providing the same information as the respective `xarray` `__repr__` methods. `_repr_html_` methods would be a cherry on top.

iohub/ndtiff.py
```diff
@@ -101,16 +189,18 @@ def str_channel_axis(self) -> bool:
         """Channel axis is string-valued"""
         return self._str_channel_axis
 
-    def _check_coordinates(
-        self, p: Union[int, str], t: int, c: Union[int, str], z: int
-    ):
+    @property
+    def ndtiff_axes(self) -> tuple[str]:
+        return ("position", "time", "channel", "z", "y", "x")
```
ieivanov, line 194: Should this be a private property? 

NDTiff allows for arbitrary axis names. Does this mean that we are not supporting axis names outside of the ones defined here? That would be OK with me.

ziw-liu: I didn't know that these can be arbitrary strings. Then how can we support that if e.g. the channel axis is not named 'channel'?

ziw-liu:
> It would also be very helpful to include examples of the new slicing and string-indexing that xarray enables. Most/all of the official iohub examples currently focus on the zarr reader/writer.

All of the example scripts are about NGFF or CC. That's because these are the formats we can write with iohub so test data doesn't need to be downloaded during documentation building. We can ask developers to always run pytest before building docs so the test data can be reused. But this will likely be a fragile hack.

@JoOkuma any ideas?

ieivanov:
Access to the MM summary metadata should be exposed for `NDTiffDataset` and `MMStack`. Currently, the summary metadata is available through the `mm_meta` property (which holds all MM metadata). I think `MicroManagerFOVMapping
` should include a property `summary_metadata` which returns `mm_meta['Summary']`. `MicroManagerFOVMapping.mm_meta` should probably also be renamed to `micromanager_metadata`.

iohub/ndtiff.py
ieivanov: Yeah, NDTiff would allow you to create for example `("position", "time", "lf_channels", "fl_channels", "z", "y", "x")`. I think it's OK if we don't support that. I was mostly arguing for `ndtiff_axes` to be a private property, i.e. `_ndtiff_axes`. We could check that the ndtiff dataset axes are in `_ndtiff_axes` and throw a "not supported" error if that's not the case.

ziw-liu: https://github.com/czbiohub-sf/iohub/pull/185/commits/a6e13a8ae2a76d8e6e02afec4ad91d0debbe59f8

ziw-liu:
@ieivanov I opened #212. Let's handle that in a new PR.

mattersoflight:
@ziw-liu fantastic work across all the layers of `iohub`. It looks like some of the open issues (#71 , #207 , #208 ) are quick to handle. Please test/fix the code if the original authors are busy and close them. 

I suggest merging when these features are implemented and tested with real data:
* iohub convert can chunk by z (looks like that was done in another PR).
* Unified API is documented on README so that power users/reviewers begin to switch and can provide feedback aligned with future development.

I think the following can be future/smaller PRs that immediately follow this:
* Proper metadata parsing across the formats we use requires a lot more thought. How we handle metadata depends on how we will consume metadata - in the viewer, through API, etc.  One PR per format would make sense to iterate through the metadata handling systematically. 
* User friendly examples.

I don't have input on when it is best to rename methods.  Since many users are now finding this PR valuable and will provide input as they use the code, it is smart not to delay the merging.

ziw-liu:
@ieivanov Is there any blocking issues left before we are ready to merge this?

ieivanov:
One question / request before I approve - when are the [docs](https://czbiohub-sf.github.io/iohub/main/api.html) rebuilt? It looks like only on a push to `main`? It will be very helpful to have docs built for this PR (or the `ometiff-uapi` branch) so that users can have easy access to the updated API.

ziw-liu:
> One question / request before I approve - when are the [docs](https://czbiohub-sf.github.io/iohub/main/api.html) rebuilt? It looks like only on a push to `main`? It will be very helpful to have docs built for this PR (or the `ometiff-uapi` branch) so that users can have easy access to the updated API.

Currently it is only built when pushing to main. I can attempt to also build for this branch, but we won't know if it is going to work (before we actually merge this).

ziw-liu:
@ieivanov I propose that we handle the docs preview in a separate PR so we can easily revert the change after the transition.

ieivanov (approved):
Let's get this PR merged. A few outstanding non-blocking issues from my point of view based on our discussion so far:

- [ ] I suggest renaming classes as in https://github.com/czbiohub-sf/iohub/pull/185#issuecomment-1955724023. This does not introduce any use-facing changes, but I think would improve the readability of the code.
- [ ] build docs preview showing the new UAPI, to be resolved in a different PR

JoOkuma:
> `stack` could be replaced with `dataset`, if you prefer. One more iteration could be:

I like @ieivanov's suggestion; I think `dataset` would be more appropriate because `stack` is a "well-defined" data structure, which is not what we use for these classes.

JoOkuma (commented):
Once again, I would like to thank @ziw-liu for this PR. It's a huge change and will improve our implementation of downstream image processing/analysis routines.

It closes the Zarr and file issues we raised before.

My concerns are:
 
- The primary usage of `xdata` on the new FOV classes, which deviates from the common API. From what I chatted with @ziw-liu, this is a requirement for the TIFF data (so it has array-like behavior? I don't remember quite well).
    I'm worried some routines start depending on it and, therefore, are not universal, which is the main goal of this PR.
   This method could be added to `BaseFOV` so every class must implement its own `xdata` method; the downside is `BaseFOV` starts deviating slightly from being an `ArrayLike` object.
- The convert function being TIFF specific. The original vision of the universal API was that the conversion function would be `BaseFOVMapping -> BaseFOVMapping` --- despite, in practice, the output type being a subset of all supported types because some are read-only.
    This leads me to think that the API isn't universal yet.
    I understand that MicroManager-specific features are not available in other classes, but I think these could be wrapped into existing `BaseFOVMapping` methods; if that's not the case, the `BaseFOVMapping` could be improved.

iohub/convert.py
```diff
@@ -1,33 +1,27 @@
...
-)
+from iohub.reader import MMStack, NDTiffDataset, read_images
 
 __all__ = ["TIFFConverter"]
+_logger = logging.getLogger(__name__)
+
 MAX_CHUNK_SIZE = 500e6  # in bytes
 
 
 def _create_grid_from_coordinates(
```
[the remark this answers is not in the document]
ziw-liu: Thanks for the catch! This is actually a bug. Fixed in https://github.com/czbiohub-sf/iohub/pull/185/commits/e8fe720519788cb8585a71d5fb54bbe7c2ffe3ad. Now this file is 95% covered.

ziw-liu:
@JoOkuma Can #85 be closed by this?

ziw-liu:
>The primary usage of xdata on the new FOV classes, which deviates from the common API. From what I chatted with @ziw-liu, this is a requirement for the TIFF data (so it has array-like behavior? I don't remember quite well).
I'm worried some routines start depending on it and, therefore, are not universal, which is the main goal of this PR.
This method could be added to BaseFOV so every class must implement its own xdata method; the downside is BaseFOV starts deviating slightly from being an ArrayLike object.

`xarray.DataArray` is also `ArrayLike`. In the case of TIFF readers it is only to label the axes of the underlying `dask.Array`. I think the spirit of UAPI was that one can call `napari.add_image` with any `BaseFOV` child.

JoOkuma:
> @JoOkuma Can #85 be closed by this?

@ziw-liu , I think so.

JoOkuma:
> > The primary usage of xdata on the new FOV classes, which deviates from the common API. From what I chatted with @ziw-liu, this is a requirement for the TIFF data (so it has array-like behavior? I don't remember quite well).
> > I'm worried some routines start depending on it and, therefore, are not universal, which is the main goal of this PR.
> > This method could be added to BaseFOV so every class must implement its own xdata method; the downside is BaseFOV starts deviating slightly from being an ArrayLike object.
> 
> `xarray.DataArray` is also `ArrayLike`. In the case of TIFF readers it is only to label the axes of the underlying `dask.Array`. I think the spirit of UAPI was that one can call `napari.view_image` with any `BaseFOV` child. Edit: IIRC we decided that the full array interface is too broad to implement.

Ok, now that I applied closer attention I understand that you're mainly using `xdata` inside the `__getitem__` methods.
I was a bit concerned because of 

```python
to_zarr(fov.xdata.data.rechunk(self.chunks), zarr_img)
```

But there isn't an easy way around it.

@ziw-liu , I'm very happy with the PR, and for me it's ready to merge and remaining tasks could be converted to issues.

ziw-liu:
@ieivanov @JoOkuma @talonchandler Thanks for the reviews!

Issue #184, opened 2023-09-01
iohub rechunk CLI


mattersoflight:
During the DL course, @edyoshikun needed to re-chunk the OME-Zarr store to improve the data i/o efficiency during training.  

Ed, please turn your code into an `iohub rechunk <store>.zarr` CLI that converts the dataset, instead of overwriting it. 

Relevant tool for zarr: https://rechunker.readthedocs.io/en/latest/

edyoshikun:
This is what I was using to do the cropping and rechunking. Two functions that might be useful and we can discuss wether they should be utilities that live within a specific project or in iohub. 

```python
# %%
from iohub.ngff import open_ome_zarr, Position, TransformationMeta
import numpy as np
import multiprocessing as mp
from natsort import natsorted
import glob
from pathlib import Path
import click
from functools import partial
import itertools
from typing import Tuple

input_data_path = "/hpc/projects/comp.micro/mantis/2023_08_09_HEK_PCNA_H2B/xx-mbl_course_H2B/cropped_dataset_v3.zarr/*/*/*"
output_data_path = "./cropped_dataset_v3_small.zarr"


def get_output_paths(input_paths: list[Path], output_zarr_path: Path) -> list[Path]:
    """Generates a mirrored output path list given an input list of positions"""
    list_output_path = []
    for path in input_paths:
        # Select the Row/Column/FOV parts of input path
        path_strings = Path(path).parts[-3:]
        # Append the same Row/Column/FOV to the output zarr path
        list_output_path.append(Path(output_zarr_path, *path_strings))
    return list_output_path


def create_empty_zarr(
    position_paths: list[Path],
    output_path: Path,
    output_zyx_shape: Tuple[int],
    chunk_zyx_shape: Tuple[int] = None,
    voxel_size: Tuple[int, float] = (1, 1, 1),
) -> None:
    """Create an empty zarr store mirroring another store"""
    DTYPE = np.float32
    MAX_CHUNK_SIZE = 500e6  # in bytes
    bytes_per_pixel = np.dtype(DTYPE).itemsize

    # Load the first position to infer dataset information
    input_dataset = open_ome_zarr(str(position_paths[0]), mode="r")
    T, C, Z, Y, X = input_dataset.data.shape

    click.echo("Creating empty array...")

    # Handle transforms and metadata
    transform = TransformationMeta(
        type="scale",
        scale=2 * (1,) + voxel_size,
    )

    # Prepare output dataset
[...]

PR #186, opened 2023-09-10
Cli rechunking

edyoshikun:
This was some of the code used during the DL course for rechunking #184

### DIFF
--- a/iohub/cli/rechunk.py
+++ b/iohub/cli/rechunk.py
@@ -1,0 +1,167 @@
+import copy
+import itertools
+import json
+import logging
+import multiprocessing as mp
+import os
+from functools import partial
+from glob import glob
+from pathlib import Path
+from typing import Literal, Tuple, Union
+
+import numpy as np
+from dask.array import to_zarr
+from natsort import natsorted
+from numpy.typing import NDArray
+from tqdm import tqdm
+from tqdm.contrib.itertools import product
+
+from iohub.ngff import Plate, Position, TransformationMeta, open_ome_zarr
+
+MAX_CHUNK_SIZE = 500e6  # in bytes
+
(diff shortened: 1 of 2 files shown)


edyoshikun:
For now, I implemented some functions that help me get to registered, rechunked and cropped volumes. Such functions work with slurm, which are beneficial for our mantis datasets with >100Positions multiple timepoints. 

These are in the standard we set of ZYX and return a ZYX volume like our existing functions (i.e deskewing, apply_affine, etc) in the mantis repo [here](https://github.com/czbiohub-sf/mantis/pull/102/files#diff-1093b95a67b463e180ba925108ceedc9cb8eb32d137152e85b5468b02c94ed62)

I don't think these should live in iohub, but linking in case we grab some inspiration for when this is implemented.

iohub/cli/rechunk.py
```diff
@@ -0,0 +1,169 @@
...
+import numpy as np
+from dask.array import to_zarr
+from natsort import natsorted
+from numpy.typing import NDArray
+from tqdm import tqdm
+from tqdm.contrib.itertools import product
+
+from iohub.ngff import Plate, Position, TransformationMeta, open_ome_zarr
+
+MAX_CHUNK_SIZE = 500e6  # in bytes
```
JoOkuma, line 21: Hi @edyoshikun , I bumped into this because I'm having issues with chunking when copying data from Daxi. 

MAX_CHUNK_SIZE should be a constraint, except for the 2147483647 bytes limit of zarr.
Otherwise, it decreases iohub flexibility.

ziw-liu:
@edyoshikun should we close this since biahub now does this in parallel?

edyoshikun:
Closing as biahub concatenate does this

PR #189, opened 2023-09-25
Extend HCS label parsing

ieivanov:
This PR extend HSC site label parsing with the format like `1-Pos000_000`


ziw-liu:
Can you modify the failing test case to reflect the expected behavior?

ieivanov:
Can I defer this to you? I think in the `MM20_ome-tiffs` collection there are datasets that use the `1-Pos000_000` name format.

ieivanov:
The tests occasionally fail because `test_converter_ometiff_hcs_numerical` samples from `MM20_ome-tiffs` and enforces hcs format (`hcs_plate=True`) even though some datasets in that collection do not have more than one position

iohub/convert.py
```diff
@@ -307,15 +307,15 @@ def _get_pos_names(self):
...
+        try:
             self.pos_names = []
             for p in range(self.p):
                 try:
                     name = self.reader.stage_positions[p]["Label"]
                 except (IndexError, KeyError):
                     name = p
                 self.pos_names.append(name)
-        else:
+        except Exception:
```
ziw-liu, line 318: Can you enumerate the expected exceptions here?

ziw-liu: Also what happens if the dataset has multiple positions but gets caught in this branch? Should it `raise`?

ieivanov: The confusing part seems to be the nested `try` statements, which are not necessary. Simplified.

ieivanov:
@edyoshikun @ziw-liu let's try to merge this today, we'll likely need to use this branch to reconstruct the data from tomorrow's experiment. This branch will also fix the failing tests in the main branch.

ziw-liu (approved):
@edyoshikun can you take a look at the proposed behavior and confirm it does not break your workflow?

edyoshikun (approved):
This does not break my workflow since we still support the original '0/0/0' position. It can also adapt to this

PR #191, opened 2023-09-29
Fix for when ChNames key is not available

JoOkuma:
Hi @ziw-liu,

I had some issues when converting DaXi's data.
This fixes the issue.

I noticed you had an exception above for something similar, so I might be missing something, and this needs improvement.


ziw-liu:
I think the case above only accounts for MM-2.0.0beta. Feel free to improve the code duplication situation. I'm also ok to merge as-is.

JoOkuma:
I'm happy with merge as it. Thank you @ziw-liu

ziw-liu (approved):

PR #193, opened 2023-09-29
Fix product integer overflow

JoOkuma:
This `np.prod` call was overflowing when converting our data and resulting in invalid chunks.

### DIFF
--- a/iohub/convert.py
+++ b/iohub/convert.py
@@ -357,7 +357,7 @@ class TIFFConverter:
         # it's OK if a single image is larger than MAX_CHUNK_SIZE
         while (
             chunks[-3] > 1
-            and np.prod(chunks, dtype=float) * bytes_per_pixel > MAX_CHUNK_SIZE
+            and np.prod(chunks, dtype=np.int64) * bytes_per_pixel > MAX_CHUNK_SIZE
         ):
             chunks[-3] = np.ceil(chunks[-3] / 2).astype(int)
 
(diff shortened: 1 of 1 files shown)


ziw-liu (commented):
Thanks for spotting this!

Out of curiosity, what's the shape of the input shape here? `int64` should be enough for 1 billion $2048^3$ volumes.

JoOkuma:
Some DaXi volumes were approximately 3500x2048x2048 🥲

ziw-liu:
> Some DaXi volumes were approximately 3500x2048x2048 🥲

But this should still be well within the range of `int64`. Was the shape in `uint16` somehow?

iohub/convert.py
```diff
@@ -357,7 +357,7 @@ def _gen_chunks(self, input_chunks):
         # it's OK if a single image is larger than MAX_CHUNK_SIZE
         while (
             chunks[-3] > 1
-            and np.prod(chunks) * bytes_per_pixel > MAX_CHUNK_SIZE
+            and np.prod(chunks, dtype=float) * bytes_per_pixel > MAX_CHUNK_SIZE
```
ziw-liu, line 360: For example I would expect this to be sufficient for any array that can fit on our data partitions:

```suggestion
            and np.prod(chunks, dtype=np.int64) * bytes_per_pixel > MAX_CHUNK_SIZE
```

ziw-liu (approved):

PR #197, opened 2023-10-12
OpenCell reader extension

ieivanov:
The OpenCell acquisition starts with a large number of positions in the position list, but only some of those are actually imaged. When we save the data, we need to use a continuous position index starting at 0 because of a Micro-manager bug: https://github.com/micro-manager/micro-manager/issues/1723, i.e. if we are imaging positions `(5, 6, 13, 12)` in the position list, they are saved with index `(0, 1, 2, 3)`. In this case, we need to parse the correct position label from the tif filename.

I've implemented this as a special version of Micro-manager metadata. If we change the MM version on the Dragonfly microscope (unlikely) or the "raw_data" prefix (also very unlikely) this code will not work an will need to be modified. I've also switched the order of `_gather_index_maps()` and `_set_mm_meta()` to make this work. I didn't see unintended consequences here, but I think this may be making some of the tests fail. @ziw-liu can you help me take a look?

cc: @Soorya19Pradeep

### DIFF
--- a/iohub/multipagetiff.py
+++ b/iohub/multipagetiff.py
@@ -173,16 +173,21 @@ class MicromanagerOmeTiffReader(ReaderBase):
                 for ch in self.mm_meta["Summary"]["ChNames"]:
                     self.channel_names.append(ch)
 
-            # Parsing of data acquired with the OpenCell 
+            # Parsing of data acquired with the OpenCell
             # acquisition script on the Dragonfly miroscope
-            elif mm_version == "2.0.1 20220920" and self.mm_meta["Summary"]["Prefix"] == "raw_data":
-                file_names = set([(key[0], val[0]) for key, val in self.coord_map.items()])
+            elif (
+                mm_version == "2.0.1 20220920"
+                and self.mm_meta["Summary"]["Prefix"] == "raw_data"
+            ):
+                file_names = set(
+                    [(key[0], val[0]) for key, val in self.coord_map.items()]
+                )
 
                 if self.mm_meta["Summary"]["Positions"] > 1:
                     self._stage_positions = [None] * self.positions
 
                     for p_idx, file_name in file_names:
-                        site_idx = int(file_name.split('_')[-1].split('-')[0])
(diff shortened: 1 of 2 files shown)


Soorya19Pradeep:
@ieivanov , thank you so much for implementing the reader. The iohub conversion works for both opencell data as well MM data now. I have tested it on data from DragonFly OpenCell imaging and MM HCS data.

Soorya19Pradeep (approved):

ziw-liu:
> I've also switched the order of `_gather_index_maps()` and `_set_mm_meta()` to make this work. I didn't see unintended consequences here, but I think this may be making some of the tests fail. @ziw-liu can you help me take a look?

The `_set_mm_meta()` method sets `self.frames` based on MM summary metadata, which may be inaccurate for incomplete datasets, and `_gather_index_maps()` will overwrite the initial value based on the index map.

Also this line breaks single page TIFF conversion because it has an empty `reader.stage_positions`:

```py
                name = self.reader.stage_positions[p].get("Label") or p
```

iohub/multipagetiff.py
```diff
@@ -209,15 +209,17 @@ def _set_mm_meta(self):
...
-            self.slices = self.mm_meta["Summary"]["Slices"]
-            self.channels = self.mm_meta["Summary"]["Channels"]
+
+            # dimensions based on mm metadata
+            # do not reflect final written dimensions
+            # these set in _gather_index_maps
+            #
+            # self.frames = self.mm_meta["Summary"]["Frames"]
+            # self.slices = self.mm_meta["Summary"]["Slices"]
+            # self.channels = self.mm_meta["Summary"]["Channels"]
```
ieivanov, line 222: Thanks for catching this @ziw-liu. I commented out these lines so they don't overwrite the dimensions set by `_gather_index_maps`

ziw-liu (approved):
Code looks good! (I haven't tested the added functionality.)

It would be nice if this can be tested automatically somehow.

ieivanov:
Thanks @ziw-liu. I opened a new issue for adding a test dataset

PR #202, opened 2023-11-07
Modified fluorophore color code

ieivanov:
Suggesting a modification of the display color for some fluorophores.

I found that Cy5 and mCherry were both plotted in magenta, when they actually have quite different emission spectrum.

Alexa561 was also listed under both magenta and yellow colors.

### DIFF
--- a/iohub/display_utils.py
+++ b/iohub/display_utils.py
@@ -8,7 +8,15 @@ from iohub.ngff_meta import ChannelMeta, WindowDict
 """ Dictionary with key works and most popular fluorescent probes """
 CHANNEL_COLORS = {
     # emission around 510-525 nm
-    "lime": ["GFP", "Green", "Alexa488", "GCaMP", "FITC", "mNeon", "MitoViewGreen"],
+    "lime": [
+        "GFP",
+        "Green",
+        "Alexa488",
+        "GCaMP",
+        "FITC",
+        "mNeon",
+        "MitoViewGreen",
+    ],
     # emission around 580 - 610 nm
     "magenta": [
         "TXR",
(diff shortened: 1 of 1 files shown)


ziw-liu (approved):

iohub/display_utils.py
```diff
@@ -7,20 +7,22 @@
...
         "dTomato",
-        "Cy5",
         "Alexa561",
     ],
+    # emission around 440 - 460 nmm
     "blue": ["DAPI", "Blue", "BFP"],
     "red": ["Red"],
-    "orange": ["Orange", "Cy3"],
-    "yellow": ["Alexa561"],
+    "yellow": ["Cy3"],  # Cy3 emission is at 570 nm
```
edyoshikun, line 24: Since we are at it, should we add the color names as values of the color keys? i.e can we add `yellow` as an option for `yellow` and for whatever color?

edyoshikun (approved):
Otherwise these suggestions LGTM! I also noticed that on the new Opencell dataset.  Thanks @ieivanov

iohub/display_utils.py
edyoshikun: `Alexa561` also needs to be yellow then.

ieivanov:
If you ask me, I'd keep these default fluorophore colors to the bare minimum. If we start adding esoteric fluorophores, then there would be no end to this. I'd suggest removing GCaMP, mNeon, dTomato, Alexa561, Alexa568, MitoViewGreen, MitoView633, LysotrackerRed, etc. Also, isn't white the default color? In that case, I'd remove that whole key

ziw-liu:
> Also, isn't white the default color? In that case, I'd remove that whole key

Good point. We inherited the 'white' part from the legacy waveorder writer.

ieivanov:
OK, I've cut down the number of supported fluor to what I think is the bare minimum. Let me know what you think

edyoshikun:
LGTM!

PR #204, opened 2023-11-14
Save ndtiff image plane metadata per position

ieivanov:
Breaking up the ndtiff image plane metadata, which can be quite large, per position

### DIFF
--- a/iohub/convert.py
+++ b/iohub/convert.py
@@ -462,8 +462,9 @@ class TIFFConverter:
             "Converting Timepoints/Channels: |{bar:16}|{n_fmt}/{total_fmt} "
             "(Time Remaining: {remaining}), {rate_fmt}{postfix}]"
         )
-        all_ndtiff_metadata = {}
         for p_idx in tqdm(range(self.p), bar_format=bar_format_positions):
+            position_image_plane_metadata = {}
+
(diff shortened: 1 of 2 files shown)


iohub/convert.py
```diff
@@ -541,19 +542,25 @@ def _convert_ndtiff(self):
...
-            json.dump(all_ndtiff_metadata, metadata_file, indent=4)
+                    position_image_plane_metadata[frame_key] = image_metadata
+
+            logging.info("Writing ND-TIFF image plane metadata...")
+            # image plane metadata is save in
+            # output_dir/row/well/fov/img/image_plane_metadata.json,
+            # e.g. output_dir/A/1/FOV0/0/image_plane_metadata.json
+            with open(
+                os.path.join(
+                    self.output_dir, zarr_arr.path, "image_plane_metadata.json"
```
ziw-liu, line 557: I think `zarr_arr.path` is `/`-delimited regardless of the platform. This may not work on Windows.

ieivanov: Tested that this doesn't cause problems. `os` and `pathlib` correctly handle `/` as a file separator on Windows.

ziw-liu (approved):

PR #208, opened 2024-02-01
pin tifffile to a newer version to allow unpinning of ome-zarr.

ilan-theodoro:
Since version 2023.3.15, tifffile `micromanager_metadata["IndexMap"]` is stored as a numpy array instead of a dict. So right after i get `micromanager_metadata["IndexMap"]`, I am converting it again to a dict as specified in the tifffile releases before 2023.3.15.  

However, I identified in the tests that there is shape mismatch in the `test_converter.py` if `zarr<2.15` or `tifffile>=2024.1.30`, so I pinned those versions. 
https://github.com/czbiohub-sf/iohub/blob/a4d43a2a8cd8d4abce00e475f9f1699d451eba9f/tests/test_converter.py#L253

But it solves the problem in #207 and #165.


ziw-liu:
Seems like black 24.1.1 has different behavior compared to 23.1... @ilan-francisco can you change those lines so the CI passes?

ilan-francisco:
> Seems like black 24.1.1 has different behavior compared to 23.1... @ilan-francisco can you change those lines so the CI passes?

My bad, I tried to fix it before I got your message.  I will push it again

ilan-francisco:
@ziw-liu actually I discovered that not all files were used locally when I tested. So the problem persists with `tifffile=2023.3.15`. I am having a bad time trying to solve it. The problem is indeed with the dimensions of some files. For example, reading from `mm2.0-20201209_4p_2t_5z_1c_512k_1` gives different results for the versions `2023.2.28` and `2023.3.15`. Somehow reading from `tifffile=2023.3.15` kind of gives us an extra dimension.

ziw-liu:
@ilan-francisco https://github.com/cgohlke/tifffile/issues/187 gives some context. Unfortunately upstream is not interested in the particular solution I proposed.

ziw-liu:
Closing in favor of #185.

Thanks @ilan-theodoro!

PR #214, opened 2024-03-08
Fix bug determining number of rows and cols

ieivanov:
In the case where you image all columns at row 35 (i.e. 1x38 FOVs) of a 38x38 grid you'd end up with incorrect determination of the number of rows and columns as (35, 38). This PR fixes this bug.


ziw-liu (commented):
Looks good, thanks!

PR #215, opened 2024-04-02
Fix bug reading dragonfly acquisitions

ieivanov:
This PR fixes a bug in reading data acquired with the Dragonfly acquisition pipeline.

In the Dragonfly acquisition pipeline, we start with a large position list, but not all of these positions may be acquired. The resulting `ome.tif` dataset will have a large number of positions in the MM metadata, but fewer data series. In this PR, I limit the `xdata` size to the number of actual positions acquired. The site index is only accessible from the filename. There may be a better way to get the number of tif files and their filenames than doing a glob in the root directory.

### DIFF
--- a/iohub/mmstack.py
+++ b/iohub/mmstack.py
@@ -127,8 +127,6 @@ class MMStack(MicroManagerFOVMapping):
         self._store = series.aszarr()
         _logger.debug(f"Opened {self._store}.")
         data = da.from_zarr(zarr.open(self._store))
-        # not all positions in the position list may have been acquired
-        data = data[: self.positions]
         self.dtype = data.dtype
         img = DataArray(data, dims=raw_dims, name=self.dirname)
         xarr = img.expand_dims(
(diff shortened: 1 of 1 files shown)


iohub/mmstack.py
```diff
@@ -140,6 +141,8 @@ def _parse_data(self):
                 "smaller than the number of channels. "
                 f"Completing with fallback names: {self.channel_names}."
             )
+        # not all positions in the position list may have been acquired
+        xarr = xarr[: self.positions]
```
ziw-liu, line 145: Will the missing positions always be at the end of the list?

ieivanov: Just checked. Yes, this is the case, at least for now, due to a MM bug:
https://github.com/micro-manager/micro-manager/issues/1723
https://github.com/czbiohub-sf/iohub/pull/197

ziw-liu (approved):

PR #217, opened 2024-04-06
Allow undefined pixel size for NDTiff

ziw-liu:


edyoshikun (commented):
This works well for the ndtiff that are acquired from the MDA where for some reason it doesn't grab the pixel size..

I need to isolate another issue but what I noticed is that if you acquire it through the MDA using NDTiff, the channel names don't get written properly.

PR #218, opened 2024-04-17
Fix converter for ndtiff using multicam

edyoshikun:
This PR fixes handling ndtiffs written with mutlicam. When using multicam, the datasets have an extra axis that needs to be parsed to reshape the output array.

### DIFF
--- a/iohub/convert.py
+++ b/iohub/convert.py
@@ -147,6 +147,7 @@ class TIFFConverter:
             # for `get_num_positions()`
             self.p = self.reader.num_positions
         self.t = self.reader.frames
+        self.cams = self.reader.cameras  # multicam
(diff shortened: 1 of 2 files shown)


edyoshikun:
@ieivanov this is the PR I started trying to fix the convert to have iohub read the data with the additional axis.

ieivanov:
Thanks Ed - let's see if we can fix this problem at the source. I don't like that in this implementation we need to expand the array dimensions from 5D to 6D. In the meantime, I understand that this is helping you convert the data that's acquired with the intermediate fix we have.

ziw-liu:
@edyoshikun do you still need this?

edyoshikun:
I believe we fixed this at the source in MM core.

PR #221, opened 2024-05-29
DaXi dataset reader implementation

JoOkuma:
Hi, we have a new DaXi file format.
The metadata might change, so I'll leave this as a draft PR until it's fixed.


shinzlet:
@ziw-liu Unless there is code that @JoOkuma wants to extract from this for other purposes, no - DaXi has an entirely different format now (it will just be a standard ome-zarr HCS plate with some extra metadata sprinkled in)

JoOkuma:
@ziw-liu, we have stopped using it, as @shinzlet mentioned, his software saves to Zarr automatically

PR #228, opened 2024-07-02
`set-scale` utility

aaronalvarezcz:
2024-09-17 update from @talonchandler. 

**Problem:** Scale metadata in existing .zarr stores is frequently incorrect. 

**Solution:** A small CLI utility `set-scale` that can fix incorrect metadata in a single CLI call. 

**Previous work:** the `shrimPy` repository included an [earlier version of this utility.](https://github.com/czbiohub-sf/shrimPy/blob/8f2469bf779ea4efabd3cbce304c66f1dd7c834d/mantis/cli/update_scale_metadata.py). 

**Interface:**
```
iohub set-scale -i input.zarr/ -z 2.0 -y 0.325 -x 0.325
```
will result in
```
Updating all.zarr/A/1/0 z scale from 2.3 to 2.0.
Updating all.zarr/A/1/0 y scale from 2.6 to 0.325.
Updating all.zarr/A/1/0 x scale from 2.6 to 0.325.
Updating all.zarr/A/1/1 z scale from 2.3 to 2.0.
...
...
...
Updating all.zarr/D/4/8 x scale from 2.6 to 0.325.
```

**Other notes:**
- I (@talonchandler) needed to move several parsing utilities into `iohub`. These utilities are widely used across compmicro's pipelines, so I think they deserve a place here. 
- `-i input.zarr` will expand a plate into a list of positions, then update the scale metadata of each position. `-i input.zarr/*/*/*` is equivalent. 
- this utility now accepts any subset of `-t`, `-z`, `-y`, and `-x` flags


iohub/cli/cli.py
```diff
@@ -117,3 +118,18 @@ def convert(
...
+
+
+@cli.command()
+@click.help_option("-h", "--help")
+@click.argument(
+    "files",
+    nargs=-1,
+    required=True,
+    type=_DATASET_PATH,
+)
```
talonchandler, line 130: `iohub` convert uses the `-i` convention, but you're dropping it here. Are you trying to match `iohub info`?

I think it makes sense to match the convention, especially because we want the wildcard behavior from the `-i`.

talonchandler (changes requested):
I'll suggest:
- reviving the `-i` option
- trialing a full CLI and/or config interface instead of the current interactive 
- documenting and testing

ziw-liu:
Please add a description of the feature you are implementing and the problem it solves.

iohub/cli/cli.py
```diff
@@ -117,3 +119,32 @@ def convert(
...
+@click.option(
+   "--z-scale",
+   "-z",
+    required=False,
+    type=float,
+    help="New z scale",
+)
+def update_scale_metadata(input_position_dirpaths, z_scale, y_scale, x_scale):
+    """Update scale metadata in OME-Zarr datasets"""
```
talonchandler, line 149: I will suggest adding an example call like `>> iohub update-scale-metadata -i <input> ... etc` here. 

Double check that it works.

talonchandler:
After discussion, here's a list of TODOs before we're ready for wider review:

- remove the dependency on mantis and copy over the `-i` option instead.
- move the test from shrimpy over to iohub
- make sure that tests are passing

talonchandler:
I've revised this PR and added tests. It's now ready for review.

talonchandler (approved):

talonchandler:
When this merges, we can remove it from shrimpy by merging [this PR](https://github.com/czbiohub-sf/shrimPy/pull/142).

edyoshikun (approved):
LGTM! just tested this and worked well.

tests/cli/test_cli.py
```diff
@@ -103,3 +105,33 @@ def test_cli_convert_ome_tiff(grid_layout, tmpdir):
     result = runner.invoke(cli, cmd)
     assert result.exit_code == 0, result.output
     assert "Converting" in result.output
+
+
+def test_cli_update_scale_metadata():
```
ziw-liu, line 110: I recommend testing both the CLI (end-to-end testing) and the underlying API (unit testing).

iohub/cli/update_scale_metadata.py
```diff
@@ -0,0 +1,30 @@
...
+
+def update_scale_metadata(
+    input_position_dirpaths: List[str],
+    z_scale: float,
+    y_scale: float,
+    x_scale: float,
+):
+    for input_position_dirpath in input_position_dirpaths:
+        with open_ome_zarr(
+            input_position_dirpath, layout="fov", mode="a"
```
ziw-liu, line 15: Should this be `r+`?

iohub/cli/update_scale_metadata.py
```diff
@@ -0,0 +1,30 @@
...
+        ) as input_dataset:
+            print(
+                f"Updating {input_position_dirpath} scale from "
+                f"{tuple(input_dataset.scale[-3:])} to "
+                f"{z_scale, y_scale, x_scale}."
+            )
+            input_dataset.zattrs["old_scale"] = input_dataset.scale[2:]
```
ziw-liu, line 22: This should be namespaced, e.g. contained in "iohub".

iohub/cli/update_scale_metadata.py
```diff
@@ -0,0 +1,30 @@
+from typing import List
+
+from iohub import open_ome_zarr
+from iohub.ngff.models import TransformationMeta
+
+
+def update_scale_metadata(
+    input_position_dirpaths: List[str],
```
ziw-liu, line 8: Should we also allow T?

tests/cli/test_cli.py
ziw-liu: More concrete example: the function can be tested with a single FOV, and the CLI can be tested with a wild card pattern.

tests/cli/test_cli.py
```diff
@@ -103,3 +105,33 @@ def test_cli_convert_ome_tiff(grid_layout, tmpdir):
...
+    position_path = hcs_ref / "B" / "03" / "0"
+    with open_ome_zarr(position_path, layout="fov", mode="a") as input_dataset:
+        old_scale = input_dataset.scale
```
ziw-liu, line 113: This will permanently modify the reference file for the developers. See the NGFF tests for how to create a temporary copy.

ziw-liu: If you already ran this locally the pytest cache needs to be downloaded again.

iohub/cli/update_scale_metadata.py
talonchandler: Good call. Thanks!

tests/cli/test_cli.py
talonchandler: Thanks @ziw-liu, I've revised this branch to include a CLI and API with tests on both.

iohub/cli/update_scale_metadata.py
talonchandler: Thanks, now namespaced to `iohub`.

tests/cli/test_cli.py
talonchandler: Sorry for polluting the fixtures! 

Thanks for flagging. Now fixed with your `_temp_copy`.

ieivanov:
It's great that the `OptionEatAll` is now moving it `iohub`. It'll be good to add tests for it, and also test that `-i input.zarr` gets expanded into positions.

talonchandler:
@ieivanov I've added tests for `OptionEatAll` and plate-expansion behavior. I think this is ready for a final look.

ieivanov:
I've run out of time debugging, but I managed to fix an issue with updating the scale transform when the transform did not exist in the first place (datasets are created with identity transform by default).

I'm still seeing an issue that in the .zattrs file only `"old_x": 1.0` is written even when updating all of X, Y, and Z.

ieivanov (approved):
Let's merge!

PR #229, opened 2024-07-02
Addition of cropping.py

alishbaimran:
Sample script to index into either time, channels, z, y, and x into an existing ome-zarr store and create a new ome-zarr store with crops. Code is currently structured as a class with some helper methods but can be re-factored as some utils methods that are used in the script directly (without the class structure).

### DIFF
--- a/iohub/cropping.py
+++ b/iohub/cropping.py
@@ -61,23 +61,21 @@ class OMEZarrHandler:
     def slice_and_assign(self):
         print("Starting slice and assign process")
         
-        # Open the base OME-Zarr store
+        # Open OME-Zarr store
         ds = self.open_zarr_store(self.base_path)
         
         # Get positions and channel indices
         positions = self.get_positions(ds)
         channel_indices = self.get_channel_indices(ds)
 
-        # Set shape, chunks, and scale based on an example position
+        # Set shape, chunks, and scale 
         example_position = positions[0][1]  
         array_shape = example_position['0'].shape
         scale = example_position.scale
         shape = self.calculate_shape(array_shape)
         chunks = (1, 1, 1, shape[3], shape[4])
 
-        #scale = (1.0, 1.0, 0.23947195669052998, 0.39464974089282584, 0.39464974106189504)
-
-        # Generate position keys
+        # position keys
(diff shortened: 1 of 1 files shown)


edyoshikun (commented):
As discussed in our meeting with @talonchandler. I think the convenience functions like `get_positions` and`get_channel_indices` would be great to have inside iohub.
@ziw-liu what would be a good place to have these (iohub.ngff?)

iohub/cropping.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # Set shape, chunks, and scale 
+        example_position = positions[0][1]  
+        array_shape = example_position['0'].shape
+        scale = example_position.scale
+        shape = self.calculate_shape(array_shape)
+        chunks = (1, 1, 1, shape[3], shape[4])
+
+        # position keys
+        position_keys = [self.validate_position_key(path) for path, _ in positions]
```
edyoshikun, line 79: could run everything in a single for loop

ziw-liu:
> As discussed in our meeting with @talonchandler. I think the convenience functions like `get_positions` and`get_channel_indices` would be great to have inside iohub. @ziw-liu what would be a good place to have these (iohub.ngff?)

Can you explain why is `list(plate.positions())` needed as a separate API?
Also note that [`NGFFNode.get_channel_index()`](https://czbiohub-sf.github.io/iohub/main/api/ngff.html#iohub.ngff.NGFFNode.get_channel_index) exists.

JoOkuma:
Feel free to ignore this comment because no one asked for my review.

Since the unified API was merged, I think it would be nice for this function to take a `BaseFOVMapping` as input and crop to HCS ome-Zarr. This could help crop any format without converting to ome-zarr and kind of "future proof".

ziw-liu:
> Feel free to ignore this comment because no one asked for my review.
> 
> Since the unified API was merged, I think it would be nice for this function to take a `BaseFOVMapping` as input and crop to HCS ome-Zarr. This could help crop any format without converting to ome-zarr and kind of "future proof".

The catch here is that OME-Zarr UAPI is still pending...

ziw-liu:
Replaced by `biahub concatenate`.

PR #230, opened 2024-07-03
Rename_wells initial implementation

josephschull:
Initial implementation of rename_wells function in ngff.plate class. Takes in csv file with old and new well paths.

### DIFF
--- a/iohub/ngff.py
+++ b/iohub/ngff.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import logging
 import math
 import os
+import csv
 from copy import deepcopy
 from typing import TYPE_CHECKING, Generator, Literal, Sequence, Union
 
@@ -12,6 +13,7 @@ import zarr
 from numcodecs import Blosc
 from numpy.typing import ArrayLike, DTypeLike, NDArray
 from pydantic import ValidationError
+import zarr.storage
 from zarr.util import normalize_storage_path
 
 from iohub.display_utils import channel_display_settings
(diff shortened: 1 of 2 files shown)


iohub/ngff.py
```diff
@@ -1564,6 +1566,61 @@ def positions(self) -> Generator[tuple[str, Position], None, None]:
         for _, well in self.wells():
             for _, position in well.positions():
                 yield position.zgroup.path, position
+    
+    def rename_wells(self, well_names_csv: str):
```
talonchandler, line 1570: I think this signature should be `rename_well(self, old: str, new: str):`. See Ziwen's `rename_channel` for an example of what we have in mind here.

iohub/ngff.py
```diff
@@ -1564,6 +1566,61 @@ def positions(self) -> Generator[tuple[str, Position], None, None]:
...
+        rename_wells will adjust well paths from 0/0 --> 0/A, 0/1  -> 0/B
+
+        """
+        
+        names = []
+
+        with open(well_names_csv, mode='r', newline='', encoding='utf-8') as csvfile:
+            csvreader = csv.reader(csvfile)
+            for row in csvreader:
+                names.append([row[0], row[1]])
```
talonchandler, line 1592: Instead of taking a csv into this function, the CLI can handle the csv. So this block can move up to the CLI level. 

I'm imagining that the new signature of this function will let you  modify a single well via python, and the CLI level will let you rename many wells with a csv.

talonchandler: It'll be worthwhile learning how to use click to build a CLI because you'll need to do the same for zebrapro.

iohub/ngff.py
```diff
@@ -1564,6 +1566,61 @@ def positions(self) -> Generator[tuple[str, Position], None, None]:
...
+        modified_wells = []
+        modified_columns = []
+        modified_rows = []
+
+        for old_well_path, new_well_path in names:
+            old_row, old_column = old_well_path.split('/')
+            new_row, new_column = new_well_path.split('/')
+
+            for well in self.metadata.wells:
+
```
talonchandler, line 1605: Double check that you have `black` installed. I think it will squash this whitespace, and I think that one's of the reasons CI isn't passing.

iohub/ngff.py
```diff
@@ -1564,6 +1566,61 @@ def positions(self) -> Generator[tuple[str, Position], None, None]:
...
+        modified_wells = []
+        modified_columns = []
+        modified_rows = []
+
+        for old_well_path, new_well_path in names:
+            old_row, old_column = old_well_path.split('/')
+            new_row, new_column = new_well_path.split('/')
+
+            for well in self.metadata.wells:
```
talonchandler, line 1604: The for loop will move to the CLI level.

iohub/ngff.py
```diff
@@ -1564,6 +1566,61 @@ def positions(self) -> Generator[tuple[str, Position], None, None]:
...
+
+                if well.path == old_well_path and well not in modified_wells:
+                    well.path = new_well_path ### update well metadata
```
talonchandler, line 1607: Great! I think these lines (together with the lines for rows and columns) are the core functionality that will stay here.

iohub/ngff.py
```diff
@@ -1564,6 +1566,61 @@ def positions(self) -> Generator[tuple[str, Position], None, None]:
...
+        with open(well_names_csv, mode='r', newline='', encoding='utf-8') as csvfile:
+            csvreader = csv.reader(csvfile)
+            for row in csvreader:
+                names.append([row[0], row[1]]) 
+
+        names = np.array(names)  ### names is a 2D array of [old_well_path, new_well_path
+
+        modified_wells = []
+        modified_columns = []
+        modified_rows = []
```
talonchandler, line 1598: I think this handling should go to the CLI level. 

Are you saying "skip writing a well name if you've already written to it"? I think this case should live at the CLI level and should print something like "warning: multiple wells were renamed to "B/2"  

Do you handle the case where the well doesn't exist. Maybe this should raise a `ValueError` at the API level (for example, see how Ziwen handles this case in `rename_channel` where `get_channel_index` raises `ValueError` if the name does not exist. 

When this raises a `ValueError`, then the CLI level can catch the exception, print a message to the user, and continue.

talonchandler (changes requested):
Great Joseph! 

I think you've got the right idea, but I think the API level should mirror the existing `rename_channel` and the CLI level should handle the CSV file while making API calls. 

More detailed comments below.

PR #232, opened 2024-07-07
`rename-wells` CLI utility

josephschull:
This PR adds:
- a `rename-wells` utility with signature `iohub rename-wells -i input.zarr -c well-mapping.csv`, which renames the input well names in place. The `well-mapping.csv` file has two columns with old and new names, and I have an example `well-mapping.csv` in `/iohub/docs/examples`
- an API to rename single wells via `ngff.rename_well`. The CLI uses this API.


ieivanov:
Hey @josephschull @talonchandler, I'd like to test this PR - let me know if it's ready. I have a CSV like this in mind: `/hpc/projects/intracellular_dashboard/viral-sensor/2024_06_13_SEC61_TOMM20_ZIKV_DENGUE_1/0-convert/well_map.csv` but I can modify it as it makes sense

iohub/cli/cli.py
```diff
@@ -87,3 +89,60 @@ def convert(input, output, grid_layout, chunks):
         chunks=chunks,
     )
     converter()
+
+
+@click.command()
```
talonchandler, line 94:
```suggestion
@cli.command()
```
This change will add it to the group so that you can type `iohub rename-wells`.

iohub/cli/cli.py
```diff
@@ -87,3 +89,60 @@ def convert(input, output, grid_layout, chunks):
...
+@click.argument(
+    "csvfile",
+    type=click.File("r"),
+)
+@click.argument(
+    "zarrfile",
+    type=click.Path(
+        exists=True,
+        file_okay=True,
+        dir_okay=False,
```
talonchandler, line 105:
```suggestion
        dir_okay=True,
```
A .zarr is a directory, right? I couldn't get it to work without this change. How did you test this?

iohub/cli/cli.py
```diff
@@ -87,3 +89,60 @@ def convert(input, output, grid_layout, chunks):
...
+        resolve_path=True,
+        path_type=pathlib.Path,
+    ),
+)
+def rename_wells_cli(csvfile, zarrfile):
```
talonchandler, line 110:
```suggestion
def rename_wells(csvfile, zarrfile):
```
This is the CLI name, so I think you probably want `rename_wells`

iohub/cli/cli.py
talonchandler, line 105: I'm in favor of options instead of arguments (largely to match our conventions in shrimpy).

Maybe a signature like: `iohub rename-wells -i input.zarr -c names.csv`

iohub/ngff.py
```diff
@@ -1565,6 +1565,50 @@ def positions(self) -> Generator[tuple[str, Position], None, None]:
             for _, position in well.positions():
                 yield position.zgroup.path, position
 
+    def rename_well(
+        self,
+        well,
+        old_well_path: str,
```
talonchandler, line 1571: Can you explain what `well` does here? Is this parameter necessary?

iohub/cli/cli.py
```diff
@@ -87,3 +89,60 @@ def convert(input, output, grid_layout, chunks):
...
+    for old_well_path, new_well_path in names:
+        if old_well_path not in well_paths:
+            raise ValueError(
+                f"Old well path '{old_well_path}' not found "
+                f"in the plate metadata."
+            )
+        for well in plate.metadata.wells:
+            if well.path == old_well_path and well not in modified["wells"]:
+                plate.rename_well(
+                    plate, well, old_well_path, new_well_path, modified, False
```
talonchandler, line 146: I'm running into errors here. 

`TypeError: Plate.rename_well() takes from 4 to 6 positional arguments but 7 were given`

I don't think `plate` should appear as your first argument?

talonchandler (changes requested):
Hey Joseph, I've made some suggestions. Let's test this (manually first, then with simple automated tests) before we request further input from Ivan and/or Ziwen.

iohub/ngff.py
josephschull: From my understanding, well.path = new_well_path, and the eventual call to self.dump_meta() reassigns the well metadata, so I think it is necessary to include the well itself as an argument to this function rather than simply calling zarr.storage.rename(self.zgroup._store, old_well_path, new_well_path.) which renames the directory  names within the zarr store.

ieivanov: I also think that `well` shouldn't be a argument to this function. Can you not get a reference to the well from `well = self[old_well_path]`?

iohub/ngff.py
```diff
@@ -1565,6 +1565,50 @@ def positions(self) -> Generator[tuple[str, Position], None, None]:
...
+        Parameters
+        ----------
+        old_well_path : str
+            Old name of well
+        new_well_path : str
+            New name of well
+        modified : dict, optional
+            Tracks modified wells
+        single well: bool, optional
+            True: renaming one well, False: renaming multiple wells
```
ieivanov, line 1587: `modified` and `single_well` shouldn't need to be arguments either. This function should only rename one well at a time - for example, your old_well_path is a string rather than a list of strings. To modify multiple wells, the user should call this function multiple times. How do you expect the user to interact with `modified`? Should this be an internal variable, e.g. `self._modified`?

iohub/cli/cli.py
```diff
@@ -87,3 +89,60 @@ def convert(input, output, grid_layout, chunks):
...
+    help="Path to the input Zarr file.",
+)
+@click.option(
+    "-c",
+    "--csv",
+    type=click.File("r"),
+    required=True,
+    help="Path to the CSV file containing well names.",
+)
+def rename_wells(csvfile, zarrfile):
```
ieivanov, line 110: It'll be useful if this is available as a python function (i.e. not click decorated) that can be imported and used in python scripts. The CLI would then decorate and call this function, e.g.

```python
@cli.command()
@click.help_option("-h", "--help")
@click.option(
    "-i",
    "--input",
    type=click.Path(exists=True, file_okay=True, dir_okay=False),
    required=True,
    help="Path to the input Zarr file.",
)
@click.option(
    "-c",
    "--csv",
    type=click.File("r"),
    required=True,
    help="Path to the CSV file containing well names.",
)
def rename_wells_cli(csvfile, zarrfile):
    rename_wells(csvfile, zarrfile)
```

talonchandler: I like the idea of having a python-accessible way of modifying with a csv, but I will suggest swapping the names. 

The click-wrapped function is what you type at the command line, so I would suggest swapping `rename_wells` and `rename_wells_cli`.

talonchandler (changes requested):
I just tried renaming with the following file:
```
0/0,A/1
0/1,B/1
0/2,C/1
```
and found that the utility appeared to succeed, but `iohub info` did not recognize the resulting .zarr. 

I suspect this is because this example increases the number of rows? Does the utility accidentally assume the same number of rows and columns before and after?

I will send the files before and after via a separate channel.

talonchandler:
@ieivanov @ziw-liu this is ready for review and testing. 

The latest version renames wells in place without copying, updating the metadata appropriately. It adds new rows/columns and deletes empty rows/columns from the metadata as expected. 

In addition to automated tests, I made a personal copy of Ivan's `/hpc/projects/intracellular_dashboard/viral-sensor/2024_06_13_SEC61_TOMM20_ZIKV_DENGUE_1/0-convert/2024_06_13_SEC61_TOMM20_ZIKV_DENGUE.zarr/` and tested it with **`iohub rename-wells -i plate.zarr -c names.csv`** where `names.csv` is a headerless two-column file:
```
0/1,A/1
0/2,A/2
0/3,A/3
0/4,A/4
0/5,B/4
0/6,B/3
0/7,B/2
0/8,B/1
```

This test met my standards:
- the renaming was very fast (no copying)
- `iohub info` worked correctly before and after
- no evidence of old well names in `.zattrs`
- `napari-iohub` can read and display the new well names

ieivanov:
I ran into a bad error trying to use a CSV file like this:

```
0/1, B/1
```

The extra white space cased problems. I've added changed to strip any white spaces and check for correct well names (using `PlateAxisMeta`) before making any changes to the zarr store. In the process I also refactored `rename_wells` into its own file, outside `cli.py`.

ieivanov:
Tests fails because of #253. The tests to rename well pass successfully

ieivanov (approved):

talonchandler:
Nice bug find and fix!

I like it in its own file, but do you want it in `iohub/cli/rename_wells.py` instead of `iohub/rename_wells.py`? I'm flexible.

Ready to merge from my perspective.

talonchandler (approved):

ieivanov:
>I like it in its own file, but do you want it in iohub/cli/rename_wells.py instead of iohub/rename_wells.py? I'm flexible.

The `convert` and `print_into` functions are also at the `iohub/` level. I think it's fine for now, we could organize the files a bit later. 

Similar to `set_scale`, should we write `metadata` on the prior well name?

PR #233, opened 2024-07-16
Refactor NGFF module and migrate to Pydantic v2

ziw-liu:
**(Breaking)** Moved the NGFF files to a subpackage.

Migrate to Pydantic v2. Downstream packages need to use the `v1` namespace as shown in https://github.com/mehta-lab/recOrder/pull/478/commits/13ca75e78f292c20c308d42d1b33ba0c63d1015e.

This PR is laying the ground work for the upcoming NGFF UAPI implementation. But the diff is significant enough so I'd like to get it reviewed first.


JoOkuma:
@ziw-liu should you update the dependencies to `pydantic>=2`?

ziw-liu:
> @ziw-liu should you update the dependencies to `pydantic>=2`?

@JoOkuma I'm not sure why https://github.com/czbiohub-sf/iohub/pull/233/commits/8ac39c715bb1115a51a5d9f97adc938c71d63900 doesn't show up in diff.

iohub/ngff/models.py
```diff
@@ -156,22 +183,25 @@ class TransformationMeta(MetaBase):
...
+        )
+        if self.type == "identity" and methods > 0:
             raise ValueError(
                 "Method should not be specified for identity transformation!"
             )
-        elif count > 2:
+        elif self.translation and self.scale:
             raise ValueError(
-                "Only one type of transformation method is allowed."
+                "Translation cannot be set to be both translation and scale."
```
JoOkuma, line 202:
```suggestion
                "'translation' and 'scale' cannot be provided in the same 'TransformationMeta'"
```
I think this might be more clear.

iohub/ngff/nodes.py
```diff
@@ -464,9 +462,11 @@ def get_tile_slice(
                     f"got type {type(pre_dims)}."
                 )
             for i, sel in enumerate(pre_dims):
+                if isinstance(sel, int):
+                    sel = slice(sel)
                 if sel is not None:
                     pad[i] = sel
-        return tuple(pad) + (r_slice, c_slice)
+        return tuple((pad + [r_slice, c_slice]))
```
JoOkuma, line 469:
```suggestion
        return tuple(*pad, r_slice, c_slice)
```
Personal preference, this suggestion could be ignored.

JoOkuma (commented):
Not much to add, great PR. The improved typing is much appreciated.

iohub/ngff/nodes.py
ziw-liu: I was trying to appeal pyright, as is the case with type checkers, it complains about your totally valid suggestion...

ieivanov (approved):
I think this looks good, free free to merge when you're happy with it. Somewhere could you give more details on what belongs in `ngff.models`, `ngff.nodes`, and `ngff.display`? Is there a parallel to structures in the TIFF data readers? With the Universal API, do you think we can also achieve similar code structure?

JoOkuma (approved):

JoOkuma:
@ziw-liu sorry, I forgot to approve after my review.

ziw-liu:
> Somewhere could you give more details on what belongs in ngff.models, ngff.nodes, and ngff.display?

See https://github.com/czbiohub-sf/iohub/pull/233/commits/f718d73c966bce12bd1cb0e52301ff62136f7738.

> Is there a parallel to structures in the TIFF data readers? With the Universal API, do you think we can also achieve similar code structure?

What structure are you envisioning? I organized NGFF files here so that on the top level (`from iohub import ...`) every file format gets a single namespace. Since the other formats have much less LOC I think they can be kept in a single file as is.

PR #234, opened 2024-07-16
Porting `create_empty_zarr` to iohub

edyoshikun:
This PR adds shrimPy utility functions relevant to iohub listed below:

- `create_empty_zarr`: enables the creation of an empty zarr
- `process_single_position`: function that can apply a czyx in czyx out function to a zarr store.

TODO:

- [x] Add docstring documentation to functions
- [ ] ~Decide on the processing BaseFOV vs Position level~ moved to #249 
- [x] Names of the functions
- [x] Hypothesis testing


mattersoflight:
* Implement tests via `hypothesis`

ieivanov:
Heads up, I'll have a bunch of comments and edits here, WIP

iohub/ngff_utils.py
ieivanov, line 1: May be better to place these in ngff/utils.py to match the upcoming #233

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+
+import click
+import numpy as np
+from numpy.typing import DTypeLike
+
+from iohub.ngff import Position, open_ome_zarr
+from iohub.ngff_meta import TransformationMeta
+
+
+def create_empty_hcs_zarr(
```
ieivanov, line 18: I am thinking about how this function is different from the standard `open_ome_zarr`. I think a better name for it may be `extend_hcs_zarr` since it adds positions and channels if they do not exist? This functionality is not evident from the current name `create_empty_hcs_zarr`.

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+    position_keys : list[Tuple[str]]
+        Position keys, will append if not present in the plate.
+        e.g. [("A", "1", "0"), ("A", "1", "1")]
+    shape : Tuple[int]
+    chunks : Tuple[int]
+    scale : Tuple[float]
+    channel_names : list[str]
+        Channel names, will append if not present in metadata.
+    dtype : DTypeLike
+    """
```
ieivanov, line 47: Add documentation on all parameters, particularly how we handle chunk sizes. What would happen if the there is already a zarr store at `store_path` with FOVs of different shape / chunks / dtype? A store like that is valid and `iohub` would allow you to save data like that, but viewers may be unhappy. `iohub info` reports a chunk size - what would happen if FOVs use different chunking? Should we check for that and disallow it in this function?

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+    # Create plate
+    output_plate = open_ome_zarr(
+        str(store_path), layout="hcs", mode="a", channel_names=channel_names
+    )
+
+    # Create positions
+    for position_key in position_keys:
+        position_key_string = "/".join(position_key)
+        # Check if position is already in the store, if not create it
+        if position_key_string not in output_plate.zgroup:
```
ieivanov, line 73: Use the output_place.positions() iohub API

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+            )
+        else:
+            position = output_plate[position_key_string]
+
+    # Check if channel_names are already in the store, if not append them
+    for channel_name in channel_names:
+        # Read channel names directly from metadata to avoid race conditions
+        metadata_channel_names = [
+            channel.label for channel in position.metadata.omero.channels
+        ]
```
ieivanov, line 90: Why not use the `Position.channel_names` API directly? What race condition do you anticipate?

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+    output_path: Path,
+    time_indices_in: list = "all",
+    time_indices_out: list = [],
+    input_channel_idx: list = [],
+    output_channel_idx: list = [],
+    num_processes: int = mp.cpu_count(),
+    **kwargs,
+) -> None:
+    """
+    Register a single position with multiprocessing parallelization over T and C
```
ieivanov, line 179:
```suggestion
    Apply function to data in an `iohub` `Position`, parallelizing over time and channel indices, and save result in an output Zarr store.
```

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+
+# TODO: modify how we get the time and channesl like recOrder (isinstance(input, list) or instance(input,int) or all)
+def process_single_position(
+    func,
+    input_data_path: Path,
+    output_path: Path,
+    time_indices_in: list = "all",
+    time_indices_out: list = [],
+    input_channel_idx: list = [],
+    output_channel_idx: list = [],
```
ieivanov, line 174: These should follow the same naming convention, probably `input_channel_indices`, `output_channel_indices`, etc. I would also stick to the same defaults - if providing an empty list, then all channels / timepoints will be processed. Should emtpy list or None be the default here? In the most common use case we shouldn't need to pass any of these parameters.

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+    **kwargs,
+) -> None:
+    """
+    Register a single position with multiprocessing parallelization over T and C
+
+    Parameters
+    ----------
+    func : CZYX -> CZYX function
+        The function to be applied to the data
+    input_data_path : Path
```
ieivanov, line 185: Rename to `input_position_path`? In the docstring give an exampel of `input.zarr/0/0/0`

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+        Must match time_indices_in if not empty.
+        Typically used for stabilization, which needs a per timepoint processing
+    input_channel_idx : list
+        The channel indices to process.
+        If empty list, process all channels.
+        Must match output_channel_idx if not empty
+    output_channel_idx : list
+        The channel indices to write to.
+        If empty list, write to all channels.
+        Must match input_channel_idx if not empty
```
ieivanov, line 203: What are acceptable values here? We should provide examples. See https://numpy.org/doc/stable/user/basics.indexing.html. I think acceptable values should be a list of slices, e.g. [slice(0, 2), slice(2, 4), ...], or list of lists, e.g. [[0], [1], ...]. In both cases data[input_channel_idx[0]] would return a CZYX array, i.e. single dimensions are not dropped. We should not allow for list of integers, e.g. [0, 1, ...] as data[input_channel_idx[0]] would return a ZYX array which is not compatible with downstream processing and you need to do data[c_idx:c_idx+1]. Also, we should provide examples on how 5 channels can be turned into 3 channels during biref recon, for example. In that case, the input and output channel indices would look like [[0, 1, 2, 3, 4]] and [[0, 1, 2]], correct? This syntax then allow for easy multiprocessing along members of zip(input_channel_idx, output_channel_idx).

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+        The path to input position
+    output_path : Path
+        The path to output OME-Zarr Store
+    time_indices_in : list
+        The time indices to process.
+        Must match time_indices_out if not "all"
+    time_indices_out : list
+        The time indices to write to.
+        Must match time_indices_in if not empty.
+        Typically used for stabilization, which needs a per timepoint processing
```
ieivanov, line 195: Similar as below, what are acceptable values. In the simple case these should be [0, 1, 2, ...], right? What if we want to perform averaging along time? Can we pass time_indices_in = [[0, 1, 2], [3, 4, 5], ...] and time_indices_out = [0, 1, ...]. As long as func handles that, it should be possible, right?

iohub/ngff_utils.py
ieivanov: Also, why are Z indices left out? What if we want to use this function to do 2D projection, and pass input_z_idx = [slice(None)], output_z_idx = [0]. This should be almost doable, right?

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+    non_func_args = {}
+
+    for k, v in kwargs.items():
+        if k in all_func_params:
+            func_args[k] = v
+        else:
+            non_func_args[k] = v
+
+    # Write the settings into the metadata if existing
+    if "extra_metadata" in non_func_args:
```
ieivanov, line 256: Add to docstrin that we can provide "extra_metadata"

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+        ]
+        partial_apply_transform_to_zyx_and_save = partial(
+            apply_transform_to_zyx_and_save,
+            func,
+            input_dataset,
+            output_path / Path(*input_data_path.parts[-3:]),
+            **func_args,
+        )
+    else:
+        # If C is empty, use only the range for time_indices_in
```
ieivanov, line 284: With good structure for input and output T, C, and Z indices, we should be able to simplify this to:

```python
for (t_idx_in, c_idx_in, z_idx_in), (t_idx_out, c_idx_out, z_idx_out) in zip(
    itertools.product(input_time_indices, input_channel_indices, input_slice_indices),
    itertools.product(output_time_indices, output_channel_indices, output_slice_indices),
)
    output[t_idx_out, c_idx_out, z_idx_out] = func(input[t_idx_in, c_idx_in, z_idx_in])
```

ieivanov:
P.S. I'm leaving a bunch of comments here so we can agree on the structure, but I'm also happy to implement some of these suggestions, I'm not asking others (Ed) to implement all of these.

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+        Position keys, will append if not present in the plate.
+        e.g. [("A", "1", "0"), ("A", "1", "1")]
+    shape : Tuple[int]
+    chunks : Tuple[int]
+    scale : Tuple[float]
+    channel_names : list[str]
+        Channel names, will append if not present in metadata.
+    dtype : DTypeLike
+    """
+    MAX_CHUNK_SIZE = max_chunk_size_bytes  # in bytes
```
JoOkuma, line 48: Is there any reason to use the `MAX_CHUNK_SIZE` variable directly rather than `max_chunk_size_bytes`?

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+        if channel_name not in metadata_channel_names:
+            position.append_channel(channel_name, resize_arrays=True)
+
+
+def apply_transform_to_zyx_and_save(
+    func,
+    position: Position,
+    output_path: Path,
+    input_channel_indices: list[int],
+    output_channel_indices: list[int],
```
JoOkuma, line 100: Maybe make `output_channel_indices` and `t_idx_out` optional and, by default, use the same as the input?

Why not use a `BaseFOV` rather than `Position`?

iohub/ngff_utils.py
```diff
@@ -0,0 +1,341 @@
...
+# TODO: modify how we get the time and channesl like recOrder (isinstance(input, list) or instance(input,int) or all)
+def process_single_position(
+    func,
+    input_data_path: Path,
+    output_path: Path,
+    time_indices_in: list = "all",
+    time_indices_out: list = [],
+    input_channel_idx: list = [],
+    output_channel_idx: list = [],
+    num_processes: int = mp.cpu_count(),
```
JoOkuma, line 175:
```suggestion
    num_processes: int = 1
```

`mp.cpu_count` can cause issues in the HPC if you allocate fewer cpus then the total cpus in the node -- which is often the case.

JoOkuma (commented):
The functions to `process_` and `transform_` the data could be made for general and work on arbitrary axes with some changes. We did this [here in ultrack](https://github.com/royerlab/ultrack/blob/1ad2619287e936b2da3f560de40d72aeaf761596/ultrack/utils/array.py#L96-L143)

Is it really necessary to have a function to process a single position?

From my POV, these "apply" functions should take an arbitrary `BaseFOVMapping` as input and write an ome-zarr.

We could write a wrapper of the `BaseFOV` that creates a "trivial" `BaseFOVMapping` that behaves as an HCS, so there's a single implementation for both cases and supporting multiple file types as input.

iohub/ngff_utils.py
ziw-liu: I think this could be a class method of `Plate`, like `Plate.from_positions`

ziw-liu: On second thought this could create a reference circle due to use of `open_ome_zarr`. So a separate function is also fine. In this case I would rename it to either `create_empty_plate` or `create_empty_hcs_ome_zarr` since 'HCS Zarr' is not specific.

iohub/ngff_utils.py
edyoshikun: no reason. fixing this. thanks

iohub/ngff_utils.py
edyoshikun: That should be part of the `concat`/`crop` function. Process single position funcitons operate at a CZYX level.

iohub/ngff_utils.py
edyoshikun: @ziw-liu needs to implement `Position` as a subclass of `BaseFOV`

iohub/ngff_utils.py
edyoshikun: I need to check but that should be possible. It's similar to the input_channel_indices.

iohub/ngff_utils.py
edyoshikun: Added documentation to all the parameters. I guess that if the the store already exists and `chunks` is not the same as the current one, then we can raise a warning. If you create one from scratch I think `chunks` should be exposed.

iohub/ngff_utils.py
edyoshikun: Still need to test this but I added the docstring and var types.

ieivanov:
Related: https://github.com/czbiohub-sf/shrimPy/issues/123

iohub/ngff_utils.py
edyoshikun: renaming to `create_empty_plate`

iohub/ngff_utils.py
```diff
@@ -0,0 +1,423 @@
...
+
+    click.echo(f"\nStarting multiprocess pool with {num_processes} processes")
+    with mp.Pool(num_processes) as p:
+        p.starmap(
+            partial_apply_transform_to_zyx_and_save,
+            iterable,
+        )
+
+
+def _is_nested(lst):
```
edyoshikun, line 386: test @tayllatheodoro

iohub/ngff/utils.py
```diff
@@ -0,0 +1,485 @@
...
+        The path to output OME-Zarr Store.
+    input_channel_indices : Union[list[slice], list[list[int]]]
+        The channel indices to process. Acceptable values:
+        - A list of slices: [slice(0, 2), slice(2, 4), ...].
+        - A list of lists of integers: [[0], [1], [2, 3, 4], ...].
+        If empty list, process all channels.
+    output_channel_indices : Union[list[slice], list[list[int]]]
+        The channel indices to write to. Acceptable values:
+        - A list of slices: [slice(0, 2), slice(2, 4), ...].
+        - A list of lists of integers: [[0], [1], [2, 3, 4], ...].
```
talonchandler, line 153: Is "[[0], [1], [2, 3, 4], ...]" a valid input? Can a single `func` take different numbers of channels? 

I *think* we need to specify that `func` takes a CZYX and returns a (C')ZYX, where C is the length of the inner `input_channel_indices` lists, and C' is the length of the `output_channel_indices`. Is my understanding correct?

iohub/ngff/utils.py
```diff
@@ -0,0 +1,485 @@
...
+) -> None:
+    """
+    Load a CZYX array from a Position object,
+    apply a transformation, and save the result.
+
+    Parameters
+    ----------
+    func : Callable
+        The function to be applied to the data.
+        Should take a CZYX array and return a transformed CZYX array.
```
ziw-liu, line 138: This description can be more explicitly typed as `Callable[[ArrayLike], ArrayLike]`. However reading the code it seems like `Callable[[NDArray, ...], NDArray]` is the actual behavior.

iohub/ngff/utils.py
```diff
@@ -0,0 +1,485 @@
...
+        If empty list, process all channels.
+    output_channel_indices : Union[list[slice], list[list[int]]]
+        The channel indices to write to. Acceptable values:
+        - A list of slices: [slice(0, 2), slice(2, 4), ...].
+        - A list of lists of integers: [[0], [1], [2, 3, 4], ...].
+        If empty list, write to all channels.
+    input_time_indices : int
+        The time index to process.
+    output_time_indices : int
+        The time index to write to.
```
talonchandler, line 158: These should be `input_time_index` and `output_time_index`, correct?

Will these ever be different? If you added both arguments for symmetry with `input/output_channel_indices`, then should we add `input/output_position_key` (or drop `position_key` and instead change `input/output_store_path` to `input/output_position_path`).

iohub/ngff/utils.py
```diff
@@ -0,0 +1,485 @@
...
+        if channel_name not in metadata_channel_names:
+            position.append_channel(channel_name, resize_arrays=True)
+
+
+def apply_transform_to_zyx_and_save(
+    func: Callable,
+    position_key: str,
+    input_store_path: Path,
+    output_store_path: Path,
+    input_channel_indices: Union[list[slice], list[list[int]]],
```
ziw-liu, line 124:
```suggestion
    input_channel_indices: list[slice | list[int]],
```

iohub/ngff/utils.py
```diff
@@ -0,0 +1,485 @@
...
+            position = output_plate[position_key_string]
+
+    # Check if channel_names are already in the store, if not append them
+    for channel_name in channel_names:
+        metadata_channel_names = position.channel_names
+        if channel_name not in metadata_channel_names:
+            position.append_channel(channel_name, resize_arrays=True)
+
+
+def apply_transform_to_zyx_and_save(
```
talonchandler, line 119: `apply_transform_to_czyx_and_save`?

iohub/ngff/utils.py
talonchandler: After reading closer, I'm a bit lost here. 

`input/output_channel_indices` can take non-nested `list[int]`, too, right?

It seems like the list of lists (or list of slices) types are only here to fix a slurmkit bug. If I understand correctly, I suggest we stick to the simpler `Union[list[int], slice]` type, which will be easier to understand. 

We can still handle the `slurmkit` bug as is.

iohub/ngff_utils.py
talonchandler: I think we should revisit your point here, @ieivanov.

I think @ieivanov has a case in mind where we need to apply the same transform to multiple groups of channels within the same CZYX array. For example, applying the same reconstruction to several groups of channels (i.e. a birefringence recon for the first five channels and the last five channels). Here @ieivanov is suggesting changes that would allow us to multiprocess across these reconstructions. 

For reconstructions I think this case happens quite rarely; each channel needs its own reconstruction. @ieivanov do you have different processing steps in mind?

We do apply multiple reconstructions to the same CZYX array (fluorescence to the first channel, birefringence to the last 5), but I currently handle this on the recOrder side separate reconstruction calls. Admittedly, these reconstructions could be faster through multiprocessing, but so far this is very rarely the bottleneck. 

Unless we have a good grouped multi-channel problem, I suggest we change back to `input/output_channel_indices : list | slice`. 

P.S. I do want `list`, not `int`, so that we can keep the shapes consistent as @ieivanov suggests. So for a single-channel process we should pass `input/output_channel_indices = [0]`.

iohub/ngff_utils.py
talonchandler: Although this hasn't been implemented, this is a more compelling case than channels IMO.

iohub/ngff_utils.py
talonchandler: All addressed in person.

iohub/ngff_utils.py
talonchandler: Addressed in person. We're doing `itertools` product over t and c, but not z.

iohub/ngff/utils.py
talonchandler: Fixed.

iohub/ngff/utils.py
edyoshikun: what was this for?

ziw-liu: We don't enforce type checking in iohub. But it's a nice to have for downstream developers. For example you get better LSP semantics.

iohub/ngff_utils.py
talonchandler: This function signature has changed, and it now accepts an `input_position_path: Path` instead of `position: Position`.

I also this that `Position` should be a subclass of `BaseFOV`, and I've moved this to a [separate issue](https://github.com/czbiohub-sf/iohub/issues/249).

iohub/ngff/utils.py
talonchandler: Just improved the typing and docs. Thanks @ziw-liu.

talonchandler (approved):
Ready to go from me!

PR #237, opened 2024-08-06
Support falcon datasets with latest micromanager

talonchandler:
This PR adds support for
- passing falcon's XY stage positions to om.zarr metadata (the existing pattern matching failed) 
- micromanager's [new Tile Creator prefix name format](https://github.com/micro-manager/micro-manager/pull/1897)

Tested on:
Input: `/hpc/instruments/cm.falcon/zebrafish/20240808_coro1a_mpeg_1dpf_10x/fish_with_coro1a_1/`
Output: `/hpc/projects/balla_group/imaging/zebrafish_infection/falcon/2024_08_08_coro1a_mpeg_1dpf_10x/0-zarr/timelapse1.zarr`

@ieivanov the new Tile Creator prefix format might affect OPS in this future, so I'm requesting your review. 
@dsundarraman heads up that we need to convert with this branch going forward. 

TODO:
- [x] add tests that `xy_stage = pos.get("DefaultXYStage"); stage_pos = pos.get(xy_stage)` handles all previous cases


iohub/mm_fov.py
```diff
@@ -132,19 +132,34 @@ def hcs_position_labels(self):
...
+                        pos["Label"].split("Pos-")[1].split("-")
+                        for pos in self.stage_positions
+                    ]
+                    return [
+                        ("0", col, fov.replace("_", "")) for col, fov in labels
+                    ]
+
+                except Exception:
+                    labels = [pos.get("Label") for pos in self.stage_positions]
+                    raise ValueError(
```
ziw-liu, line 158: I feel that we have made this pile of error catching too big over time. At this point a regex match at the top and throwing early is probably a better solution.

ziw-liu: And by 'too big' I mean a 20-character indentation 😅

ahillsley:
Does it make sense to remove the naming convention and replace with the new? Can we just add a new `try / except` so that it can handle both old and new?

ieivanov:
>Does it make sense to remove the naming convention and replace with the new? Can we just add a new try / except so that it can handle both old and new?

We are trying to keep this code clean, so I'd prefer fixing this issue properly rather than adding switch cases. I'll make time to test this, help from @ziw-liu or @talonchandler confirming that `xy_stage = pos.get("DefaultXYStage"); stage_pos = pos.get(xy_stage)` handles all previous cases is welcome too.

ieivanov:
Tested and ready to merge after review

ieivanov (approved):

ziw-liu (approved):
Code looks good. Thanks!

PR #242, opened 2024-09-01
adding conversion to tensorstore

edyoshikun:
This PR makes a zarray into tensorstore object

PR #244, opened 2024-09-07
Use annotation instead of field for tagged union

ziw-liu:
Use annotation instead of field for tagged union to allow `list[AxisMeta]` to work with pydantic 2.9.0.

Also removed non-standard axis which was added in anticipation of [RFC-3](https://github.com/ome/ngff/pull/239) to make the fix easier.


ziw-liu:
@edyoshikun this should fix CI failures in #242.

edyoshikun (approved):
LGTM! Let's merge this and I can test #242.

PR #246, opened 2024-09-10
Limit action runners

ziw-liu:
Tweak GitHub actions settings.


ziw-liu:
@ieivanov if you know a better way to get NDTIFF examples we can also switch to it. Low priority since this is only making the test after merging PRs slower.

ieivanov (approved):

ieivanov:
Sorry for the late review - I'll leave the implementation up to you.

Issue #247, opened 2024-09-26
get the zarr dataset size with iohub info verbose


edyoshikun:
Request:
`iohub info -v /test.zarr` 

should output in the summary the size of the dataset in GB. 
output:
```
Reading file:    test.zarr
Zarr hierarchy:
/
 └── plate
     └── 0
         ├── 0
            └── 0 (1, 6, 97, 2048, 2048) float32
=== Summary ===
Format:                  omezarr v0.4
Axes:                    T (time); C (channel); Z (space); Y (space); X (space); 
Channel names:           ['Phase3D', 'Deconvolved-Mem', 'Deconvolved-Nuc', 'BF', 'TRITC - 554~595', 'Cy5 - 635~730']
Row names:               ['plate']
Column names:            ['0']
Wells:                   1
Positions:               12
Total Size[GB]:  ((position_shape)*num_positions*dtype_bytes)/1e-9
```

ziw-liu:
Still partially pending due to https://github.com/czbiohub-sf/iohub/pull/248#issuecomment-2460540357.

PR #248, opened 2024-09-26
Add decompressed OME-Zarr dataset size to iohub info

edyoshikun:
This addresses issue #247 by adding the store size and array size in GB. This is useful and simple metadata. 

I wanted to know how much memory to request for caching datasets.

### DIFF
--- a/iohub/reader.py
+++ b/iohub/reader.py
@@ -262,11 +262,15 @@ def print_info(path: StrOrBytesPath, verbose=False):
                 print("Zarr hierarchy:")
                 reader.print_tree()
                 positions = list(reader.positions())
+                total_GB = len(positions) * (positions[0][1][0].nbytes) / 1e9
                 msgs.append(f"Positions:\t\t {len(positions)}")
                 msgs.append(f"Chunk size:\t\t {positions[0][1][0].chunks}")
+                msgs.append(f"Store Size [GB]:\t\t {total_GB:.2f}")
         else:
+            total_GB = reader["0"].nbytes / 1e9
             msgs.append(f"(Z, Y, X) scale (um):\t {tuple(reader.scale[2:])}")
             msgs.append(f"Chunk size:\t\t {reader['0'].chunks}")
+            msgs.append(f"Size [GB]:\t\t {total_GB:.2f}")
         if verbose:
             msgs.extend(
                 [
--- a/iohub/reader.py
+++ b/iohub/reader.py
@@ -262,34 +262,22 @@ def print_info(path: StrOrBytesPath, verbose=False):
                 print("Zarr hierarchy:")
                 reader.print_tree()
                 positions = list(reader.positions())
-                total_bytes_uncompressed = len(positions) * (
-                    positions[0][1][0].nbytes
-                )
-                total_bytes_compressed = len(positions) * (
-                    positions[0][1][0].nbytes_stored
+                total_bytes_uncompressed = sum(
+                    p["0"].nbytes for _, p in positions
                 )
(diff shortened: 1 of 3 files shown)


ziw-liu:
Is this meant to represent the size of disk (compressed) or size in RAM (decompressed)?

edyoshikun:
I find it more use when it's decompressed rather than compressed. We can report both if needed. I think zarr.array does `nbytes_stored`. What do you guys think?

iohub/reader.py
```diff
@@ -262,11 +262,15 @@ def print_info(path: StrOrBytesPath, verbose=False):
                 print("Zarr hierarchy:")
                 reader.print_tree()
                 positions = list(reader.positions())
+                total_GB = len(positions) * (positions[0][1][0].nbytes) / 1e9
                 msgs.append(f"Positions:\t\t {len(positions)}")
                 msgs.append(f"Chunk size:\t\t {positions[0][1][0].chunks}")
+                msgs.append(f"Store Size [GB]:\t\t {total_GB:.2f}")
```
talonchandler, line 268: Possibly "Uncompressed size (GB)"?

talonchandler (commented):
I think the uncompressed size if the most valuable. 

The reported size is the expected size, not the true size (e.g. it hasn't been filled yet or there was an error). Naming is tricky---maybe "Expected uncompressed size (GB)",  "Est. size in RAM (GB)", or "Est. size (GB)"?

edyoshikun:
ended up adding `uncompressed size [GB]`

iohub/reader.py
```diff
@@ -262,11 +262,21 @@ def print_info(path: StrOrBytesPath, verbose=False):
                 print("Zarr hierarchy:")
                 reader.print_tree()
                 positions = list(reader.positions())
+                total_GB_uncompressed = (
+                    len(positions) * (positions[0][1][0].nbytes) / 1e9
```
ziw-liu, line 266: THis would be confusing if showing 0.0 GB for <100 MB. Maybe try to mimic the behavior of `du -h`?

ziw-liu: e.g. https://web.archive.org/web/20111010015624/http://blogmag.net/blog/read/38/Print_human_readable_file_size

ziw-liu: Zarr also does this: https://zarr.readthedocs.io/en/v2.18.3/_autoapi/zarr.core.Array.html#zarr.core.Array.info

```
import zarr
z = zarr.zeros(1000000, chunks=100000, dtype='i4')
z.info
Type               : zarr.core.Array
Data type          : int32
Shape              : (1000000,)
Chunk shape        : (100000,)
Order              : C
Read-only          : False
Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
Store type         : zarr.storage.KVStore
No. bytes          : 4000000 (3.8M)
No. bytes stored   : 320
Storage ratio      : 12500.0
Chunks initialized : 0/10
```

ziw-liu:
Need to add a test case before merging.

ziw-liu:
Due to upstream issue https://github.com/zarr-developers/zarr-python/issues/2174, `nbytes_stored` will be wrong for OME-Zarr. I think we should just remove this field since the zarr devs are probably not focusing on v2 bugs now.

ziw-liu:
For example this compression ratio is clearly wrong:

```
No. bytes:               88473600 [84.4 MiB]
No. bytes stored:        419 [419 B]
```

ziw-liu (approved):

Issue #255, opened 2024-10-01
test_combine_fovs_to_hcs fails


ieivanov:
`test_combine_fovs_to_hcs` fails with 

```
self = <zarr.storage.FSStore object at 0x7f06d3a73210>, key = 'H/12/0/.zgroup'
value = b'{\n    "zarr_format": 2\n}'

    def __setitem__(self, key, value):
        if self.mode == "r":
>           raise ReadOnlyError()
E           zarr.errors.ReadOnlyError: object is read-only

/opt/hostedtoolcache/Python/3.11.10/x64/lib/python3.11/site-packages/zarr/storage.py:1463: ReadOnlyError
```

beniroquai:
Hey @ieivanov, I'm facing the same issue when performing the HCS example from the github pages docs and then try to open it in napari. Is this an ongoing issue? I have used the latest version from Github.

ziw-liu:
Hi @beniroquai this is a ome-zarr-py (which napari-ome-zarr uses under the hood) bug: https://github.com/ome/ome-zarr-py/issues/248#issuecomment-2738712765. In #283 I just stopped using that library for testing as a workaround. Unfortunately as iohub users there are only two things we can do:

1. Keep asking the ome-zarr-py developers to fix it. It is not a hard fix, but they might need more feedback to be convinced.
2. Before they fix the bug, accept that napari-ome-zarr has this limitation and name the images differently. The example you are following also explained this:
> While the NGFF specification (and iohub) allows for arbitrary names, the ome-zarr-py library and the napari-ome-zarr viewer can only load positions and arrays with name ‘0’ for the whole plate.

PR #256, opened 2024-10-12
Correctly name copied Position in `Plate.from_positions`

ieivanov:
`Plate.from_positions` takes 

```
positions : dict[str, Position]
    Dictionary where keys are destination path names ('row/column/fov')
    and values are :py:class:`iohub.ngff.Position` objects.
```

In this PR, I'm updating the name of the copied FOV in the omero metadata to its new name: `fov`


ziw-liu (commented):
Looks good! Can you add a regression test for this fix?

ieivanov:
I think it will be best to add this in the `test_combine_fovs_to_hcs` test, which we currently skip - see #255. I think it has something to do with the external `ome_zarr` `Reader` but I couldn't quite figure it out. Would you be able to fix #255 or would you suggest I add the tests elsewhere?

ziw-liu:
> I think it will be best to add this in the `test_combine_fovs_to_hcs` test, which we currently skip - see #255. I think it has something to do with the external `ome_zarr` `Reader` but I couldn't quite figure it out. Would you be able to fix #255 or would you suggest I add the tests elsewhere?

That test needs to be fixed (actually should be fixed before merging the original PR without skipping, not introducing a regression blocking other PRs like this). I'll see if I can do it.

ieivanov:
@ziw-liu have you had a chance to check why the `ome_zarr` `Reader` fails here?

ziw-liu (approved):

PR #259, opened 2024-11-19
Remove BOM from well name

ieivanov:
CSV files generated on Windows contain the `u'\ufeff'` byte order mark (BOM) character. By specifying the `utf-8-sig` encoding we remove the BOM character when reading the file.


tayllatheodoro (approved):

tayllatheodoro:
I used and it works!

PR #261, opened 2024-11-26
ImageArray Initialization

### LINKED ISSUE #226: TypeError: ImageArray.__init__() got an unexpected keyword argument 'store'
I occasionally get this error when using dask to write to iohub arrays. @ziw-liu I'll try to come up with a reproducible example. Let me know if you happen to know what the problem might be.

ieivanov:
`ImageArray` is designed to be initialized with a zarr array. This causes problems when using ImageArrays in parallel compute using dask, where arrays are initialized using key-value arguments, as in https://zarr.readthedocs.io/en/stable/_modules/zarr/core.html#Array. Here is a minimal example

```python
import dask.array as da
from iohub import open_ome_zarr
from multiprocessing import Pool

data_path = path_to_data  # path to any zarr store

with open_ome_zarr(data_path) as ds:
    data = da.from_zarr(ds.data)

func = sum  # you favorite funciton

with Pool(2) as p:
    output = p.map(func, data)
```

This example will fail, likely because of the way dask initializes the zarr arrays in the workers. In this PR, I make the zarray argument optional and allow for additional keyword arguments.

Fixed #226

### DIFF
--- a/iohub/ngff/nodes.py
+++ b/iohub/ngff/nodes.py
@@ -312,7 +312,8 @@ class ImageArray(zarr.Array):
 
     def __init__(self, zarray: zarr.Array = None, **kwargs):
         """Keyword arguments are passed to the zarr.Array constructor.
-        If a zarr.Array is provided, the constructor will use its attributes.
+        If a zarr.Array is provided, the constructor will use its attributes
+        to initialize the ImageArray.
         """
         if zarray is not None:
             kwargs.update(
--- a/iohub/ngff/nodes.py
+++ b/iohub/ngff/nodes.py
@@ -310,20 +310,24 @@ class NGFFNode:
 class ImageArray(zarr.Array):
     """Container object for image stored as a zarr array (up to 5D)"""
 
-    def __init__(self, zarray: zarr.Array):
-        super().__init__(
-            store=zarray._store,
-            path=zarray._path,
-            read_only=zarray._read_only,
-            chunk_store=zarray._chunk_store,
-            synchronizer=zarray._synchronizer,
-            cache_metadata=zarray._cache_metadata,
-            cache_attrs=zarray._attrs.cache,
-            partial_decompress=zarray._partial_decompress,
-            write_empty_chunks=zarray._write_empty_chunks,
-            zarr_version=zarray._version,
-            meta_array=zarray._meta_array,
-        )
+    def __init__(self, zarray: zarr.Array = None, **kwargs):
+        if zarray is not None:
+            kwargs.update(
+                {
+                    "store": zarray._store,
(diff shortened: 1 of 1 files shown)


iohub/ngff/nodes.py
```diff
@@ -310,20 +310,24 @@ def close(self):
...
-            cache_metadata=zarray._cache_metadata,
-            cache_attrs=zarray._attrs.cache,
-            partial_decompress=zarray._partial_decompress,
-            write_empty_chunks=zarray._write_empty_chunks,
-            zarr_version=zarray._version,
-            meta_array=zarray._meta_array,
-        )
+    def __init__(self, zarray: zarr.Array = None, **kwargs):
+        if zarray is not None:
+            kwargs.update(
```
edyoshikun, line 315:
```suggestion
            # Dynamically prepare kwargs for the base class to support Dask
            kwargs.update(
```

edyoshikun (approved):
This looks good to me. I would only add a sentence in the dosctring or comment why we added this (i.e supporting/compatibility with dask).

iohub/ngff/nodes.py
```diff
@@ -310,20 +310,24 @@ def close(self):
...
-            synchronizer=zarray._synchronizer,
-            cache_metadata=zarray._cache_metadata,
-            cache_attrs=zarray._attrs.cache,
-            partial_decompress=zarray._partial_decompress,
-            write_empty_chunks=zarray._write_empty_chunks,
-            zarr_version=zarray._version,
-            meta_array=zarray._meta_array,
-        )
+    def __init__(self, zarray: zarr.Array = None, **kwargs):
+        if zarray is not None:
```
ziw-liu, line 314: Can you explain when this method should be called without a argument? It is an internal wrapper, and if you don't need the wrapper, just doing `Position.zgroup["0"]` would give you the plain `zarr.Array` object.

ziw-liu:
Another thought on API design: if the goal is to get a convenient handle of a dask array out of an OME-Zarr array, would it be better to instead add an explicit constructor such like we did for `ImageArray.numpy()` and #242?

iohub/ngff/nodes.py
ieivanov: I added an example use case at the top and a short docstring to the `__init__` function:

```python
    def __init__(self, zarray: zarr.Array = None, **kwargs):
        """Keyword arguments are passed to the zarr.Array constructor.
        If a zarr.Array is provided, the constructor will use its attributes
        to initialize the ImageArray.
        """
```

I think this is a common enough use case, such that we should make iohub arrays work with the dask arrays in multiprocessing applications

PR #262, opened 2024-11-28
Add conversion of ImageArray to dask

ieivanov:
To replace #261


ziw-liu (approved):
LGTM, thanks!

ziw-liu:
Very minor thought: I'm debating whether `dask_array` would be a better name since dask has non-array data structures unlike numpy.

PR #263, opened 2024-12-06
More robust calculation of "z-step_um" for ndtiff datasets

ieivanov:
The calculations of `z-step_um` fails if the dataset has only a single z slice. In that case, `ZPosition_um_Intended` exists in the metadata, but the dataset does not have images at `z_idx = 1`. Here I've wrapped the calculation of `z-step_um` in a try/except statement which would catch both cases.

@ahillsley please confirm that this PR fixes your issue and approve it.

### DIFF
--- a/iohub/ndtiff.py
+++ b/iohub/ndtiff.py
@@ -140,19 +140,20 @@ class NDTiffDataset(MicroManagerFOVMapping):
         c_idx = self._ndtiff_channel_names[0]
         img_metadata = self.get_image_metadata(p_idx, 0, c_idx, 0)
 
-        pm_metadata["z-step_um"] = None
-        if "ZPosition_um_Intended" in img_metadata.keys():
+        try:
+            z0 = self.get_image_metadata(p_idx, 0, c_idx, 0)[
+                "ZPosition_um_Intended"
+            ]
+            z1 = self.get_image_metadata(p_idx, 0, c_idx, 1)[
+                "ZPosition_um_Intended"
+            ]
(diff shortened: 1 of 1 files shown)


iohub/ndtiff.py
```diff
@@ -140,19 +140,20 @@ def _get_summary_metadata(self):
...
-                    ]
-                    - self.get_image_metadata(p_idx, 0, c_idx, 0)[
-                        "ZPosition_um_Intended"
-                    ]
-                ),
-                decimals=3,
+                abs(z1 - z0), decimals=3
             ).astype(float)
+        # Will raise KeyError if dataset does not have z slices
+        # Will raise ValueError if dataset has only one z slice
```
ziw-liu, line 154: Is the difference between these two cases the existence of the Z dimension? In the converted Zarr would they both be Z=1 or would the first be empty?

ieivanov: >Is the difference between these two cases the existence of the Z dimension?

That is correct. 

>n the converted Zarr would they both be Z=1 or would the first be empty?

In both cases - no explicit z dimension, or z dimension with only one slide (rare use case), images will be written at z=0 in the converted zarr store. The change in this PR assures that `z-step_um` will be calculated only if more than one z slices are present in the dataset.

ziw-liu (approved):
Diff reads good! I'll leave the actual testing to @ahillsley.

ahillsley (approved):
Confirmed that it works on the 2024_12_06_ops0003 dataset

PR #264, opened 2024-12-10
add support for the `get_transform` analog to `set_transform`

pattonw:
given a set of scale and translation transforms, defined on both the group and individual individual images, each image will have a final scale and translation. These can now be conveniently retrieved with a little helper function that fetches all appropriate translations and applies them consecutively.

small test added to cover changes.


iohub/ngff/nodes.py
```diff
@@ -1010,6 +1010,43 @@ def get_axis_index(self, axis_name: str) -> int:
         """
         return self.axis_names.index(axis_name.lower())
 
+    def get_transforms(
```
ziw-liu, line 1013: Since this method does return get the actual transforms (as stored in the JSON files), but the accumulated effective value, I would name it more explicitly. Also does it make sense to split this into two methods so that the return type is more apparent? Something like:

```py
def get_effective_scale(self, image: str | Literal["*"]) -> TransformationMeta: ...
def get_effective_translation(self, image: str | Literal["*"]) -> TransformationMeta: ...
```

If this makes sense, we should also rewire the `scale` property to be a special case for this method to avoid duplication.

iohub/ngff/nodes.py
```diff
@@ -1010,6 +1010,43 @@ def get_axis_index(self, axis_name: str) -> int:
...
+                    transforms.extend(
+                        self.metadata.multiscales[0]
+                        .datasets[i]
+                        .coordinate_transformations
+                    )
+        elif image != "*":
+            raise ValueError(f"Key {image} not recognized.")
+
+        full_scale = np.array([1]*len(self.axes), dtype=float)
+        full_translation = np.array([0]*len(self.axes), dtype=float)
```
ziw-liu, line 1039: Likely inconsequential but vectorized:

```suggestion
        full_scale = np.ones(len(self.axes), dtype=float)
        full_translation = np.zeros(len(self.axes), dtype=float)
```

iohub/ngff/nodes.py
```diff
@@ -1010,6 +1010,43 @@ def get_axis_index(self, axis_name: str) -> int:
...
+        for transform in transforms:
+            if transform.type == "scale":
+                full_scale *= np.array(transform.scale)
+            elif transform.type == "translation":
+                full_translation += full_scale * np.array(transform.translation)
```
ziw-liu, line 1044: Does translation need to be scaled here? This would be necessary if the translation is given in pixel space, and needs to be converted into physical units. However the [spec](https://ngff.openmicroscopy.org/0.4/index.html#multiscale-md) says that they should already be in physical units (e.g. microns):

```text
It MAY contain exactly one translation that specifies the offset from the origin in physical units. If translation is given it MUST be listed after scale to ensure that it is given in physical coordinates.
```

ziw-liu (commented):
Thanks for your contribution! I've left my comments inline.

ziw-liu:
Re: CI failure: please refer to [the contributing guide](https://github.com/czbiohub-sf/iohub/blob/main/CONTRIBUTING.md#making-changes).

iohub/ngff/nodes.py
pattonw: That was my original plan, but I thought the translation was dependent on the scale. This is where my lack of familiarity with the spec shows.
I thought the translation was dependent on the scale so it made sense to do them bth at once.

iohub/ngff/nodes.py
pattonw: Ah ok. So I got it backwards. It is still a bit unclear to me what happens if the multiscale group has `[shift(5), translate(5)]` and the individual image "0" has `[shift(2), translate(2)]`. Is the effective scale 10, and the effective translation 7? That seems to make sense if its just assumed that we are never dealing with voxels, we are always in physical unit space when defining translations

pattonw:
passed the precommit checks 👍

iohub/ngff/nodes.py
```diff
@@ -1010,6 +1010,80 @@ def get_axis_index(self, axis_name: str) -> int:
...
+                self.metadata.multiscales[0].datasets
+            ):
+                if dataset_meta.path == image:
+                    transforms.extend(
+                        self.metadata.multiscales[0]
+                        .datasets[i]
+                        .coordinate_transformations
+                    )
+        elif image != "*":
+            raise ValueError(f"Key {image} not recognized.")
```
ziw-liu, line 1040: This block is duplicated between the two methods. Can you abstract it out as a shared private method?

iohub/ngff/nodes.py
```diff
@@ -1010,6 +1010,80 @@ def get_axis_index(self, axis_name: str) -> int:
         """
         return self.axis_names.index(axis_name.lower())
 
+    def get_effective_scale(
```
ziw-liu, line 1013: The `scale` property can now use this method instead of implementing its own logic.

iohub/ngff/nodes.py
```diff
@@ -1010,6 +1010,80 @@ def get_axis_index(self, axis_name: str) -> int:
...
+    ) -> TransformationMeta:
+        """Get the total coordinate scale metadata
+        for one image array or the whole FOV.
+
+        Parameters
+        ----------
+        image : str | Literal["*"]
+            Name of one image array (e.g. "0") to query,
+            or "*" for the whole FOV
+        """
```
ziw-liu, line 1025: Can you add docs for the returned value?

tests/ngff/test_ngff.py
```diff
@@ -485,6 +485,82 @@ def test_set_transform_image(ch_shape_dtype, arr_name):
...
+            ),
+        ),
+    ],
+)
+@given(
+    ch_shape_dtype=_channels_and_random_5d_shape_and_dtype(),
+    arr_name=short_alpha_numeric,
+)
+def test_get_transform_image(transforms, ch_shape_dtype, arr_name):
+    """Test `iohub.ngff.Position.set_transform()`"""
```
ziw-liu, line 541: Test needs to be updated.

JoOkuma (commented):
@ziw-liu, If `get_effective_translation[or scale]` returns a single transformation object would it be more friendly to return a tuple of floats or an array-like object?

@pattonw, I agree with the other comments from @ziw-liu

ziw-liu:
> If `get_effective_translation[or scale]` returns a single transformation object would it be more friendly to return a tuple of floats or an array-like object?

If the dataset does not have all 5 axes, what should the length of the returned sequence be? I'm guessing that @pattonw's use cases are mostly 3D instead of 5D?

pattonw:
> > If `get_effective_translation[or scale]` returns a single transformation object would it be more friendly to return a tuple of floats or an array-like object?
> 
> If the dataset does not have all 5 axes, what should the length of the returned sequence be? I'm guessing that @pattonw's use cases are mostly 3D instead of 5D?

Isn't it as simple as returning the `.scale` or `.translation` property of the `TransformationMeta`? Isn't the length always equal to the number of dimensions in the array?

iohub/ngff/nodes.py
pattonw: Yeah sorry, I meant scale not shift. I fixed it

ziw-liu:
> Isn't it as simple as returning the `.scale` or `.translation` property of the `TransformationMeta`? Isn't the length always equal to the number of dimensions in the array?

You're right! I mistakenly thought that `Position.scale` behaves differently (I was thinking of [this one](https://czbiohub-sf.github.io/iohub/main/api/mm_ometiff_reader.html#iohub.mmstack.MMStack.zyx_scale)).

tests/ngff/test_ngff.py
ziw-liu: In the spirit of unit testing, can this be separate for the two methods? Also the docstring needs to be updated.

iohub/ngff/nodes.py
```diff
@@ -969,19 +969,7 @@ def scale(self) -> list[float]:
...
-        for trans in transforms:
-            if trans.type == "scale":
-                if len(trans.scale) != len(scale):
-                    raise RuntimeError(
-                        f"Length of scale transformation {len(trans.scale)} "
-                        f"does not match data dimension {len(scale)}."
-                    )
-                scale = [s1 * s2 for s1, s2 in zip(scale, trans.scale)]
-        return scale
+        return self.get_effective_scale("*").scale
```
ziw-liu, line 972: This used to be the scale of the highest resolution, so it should be the first array instead of `"*"`.

ziw-liu: I think this is why tests are failing for https://github.com/czbiohub-sf/iohub/pull/264/commits/6fc5ad6616eeccc49d14dd84c60d616429abc9e3.

pattonw:
Separated into 2 unit tests, return plain list of floats instead of `TransformationMeta` classes, and removed the error I introduced in `scale` property to fetch the effective scale of the first dataset instead of just the root.

tests/ngff/test_ngff.py
```diff
@@ -485,6 +485,100 @@ def test_set_transform_image(ch_shape_dtype, arr_name):
...
+        (saved, target)
+        for saved, target in zip(input_transformations, target_scales)
+    ],
+)
+@given(
+    ch_shape_dtype=_channels_and_random_5d_shape_and_dtype(),
+    arr_name=short_alpha_numeric,
+)
+def test_get_effective_scale_image(transforms, ch_shape_dtype, arr_name):
+    """Test `iohub.ngff.Position.set_transform()`"""
```
ziw-liu, line 540:
```suggestion
    """Test `iohub.ngff.Position.get_effective_scale()`"""
```

tests/ngff/test_ngff.py
```diff
@@ -485,6 +485,100 @@ def test_set_transform_image(ch_shape_dtype, arr_name):
...
+        (saved, target)
+        for saved, target in zip(input_transformations, target_translations)
+    ],
+)
+@given(
+    ch_shape_dtype=_channels_and_random_5d_shape_and_dtype(),
+    arr_name=short_alpha_numeric,
+)
+def test_get_effective_translation_image(transforms, ch_shape_dtype, arr_name):
+    """Test `iohub.ngff.Position.set_transform()`"""
```
ziw-liu, line 567:
```suggestion
    """Test `iohub.ngff.Position.get_effective_translation()`"""
```

pattonw:
fixed test docstring and fixed case that was failing testing

ziw-liu (approved):
Looks great! Thanks @pattonw for your contribution!

ziw-liu:
Ready to merge if @JoOkuma doen't have further comments.

JoOkuma:
@ziw-liu, I don't have any additional comments.
Thanks for the contribution @pattonw

Issue #267, opened 2025-01-30
What to do when no metadata?


ahillsley:
I have an image with little/no metadata, but i would still like to convert it to a zarr store. Could we adjust the if statements in `iohub/iohub/mmstack.py` to still convert the image even if metadata is not present? Rather than just leave it out, the current code fails with "KeyError ___" if the metadata is not present.

I can fix this later, but don't have time rn.

PR #268, opened 2025-01-30
Fix reading Dragonfly non-OpenCell datasets with incomplete metadata

ziw-liu:
Use empty defaults when position, channel, and prefix metadata is not available.

Seems to fix https://github.com/czbiohub-sf/iohub/issues/267.


ieivanov (approved):

ziw-liu:
@ahillsley Can you test with your dataset?

ieivanov:
I tested with dataset that would otherwise throw error

```
...
    and self._mm_meta["Summary"]["Prefix"] == "raw_data"
KeyError: 'Prefix'
```

PR works well, merging.

PR #276, opened 2025-03-05
Compare MM FOVs

### LINKED ISSUE #157: Root path of BaseFOV

ziw-liu:
Make MM FOVs special cases when comparing, since they do not have individual paths.

Resolve #157.

### DIFF
--- a/tests/mmstack/test_mmstack.py
+++ b/tests/mmstack/test_mmstack.py
@@ -92,3 +92,12 @@ def test_fov_getitem(ome_tiff):
         for ch in fov.channel_names:
             assert img.sel(T=0, Z=0, C=ch, Y=0, X=0) >= 0
     mmstack.close()
+
+
+def test_fov_equal(ome_tiff):
+    ref1 = MMStack(ome_tiff)
+    ref2 = MMStack(ome_tiff)
+    for (_, fov1), (_, fov2) in zip(ref1, ref2):
+        assert fov1 == fov2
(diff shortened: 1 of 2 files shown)


tests/mmstack/test_mmstack.py
```diff
@@ -92,3 +92,20 @@ def test_fov_getitem(ome_tiff):
...
+    ref1.close()
+    ref2.close()
+
+
+def test_fov_not_equal(ome_tiff):
+    with MMStack(ome_tiff) as mmstack:
+        if len(mmstack) < 2:
+            # pass
+            return
+        assert mmstack["0"] != mmstack["1"]
```
ieivanov, line 111: Is it OK to hardcode the position names here?

ieivanov (approved):

tests/mmstack/test_mmstack.py
ziw-liu: I think that's just indices and it did pass for all the test sets we have.

PR #280, opened 2025-03-07
Check overwrite path

ziw-liu:
Implement https://github.com/mehta-lab/waveorder/issues/439#issuecomment-2706996852.

By default, disallow overwriting paths that do not contain `.zarr` to reduce data loss incidents caused by erroneous user input.

This can be disabled to restore spec conformance.


ieivanov (approved):
Looks good, thanks! I expanded the tests slightly

PR #283, opened 2025-03-19
Fix test combine FOVs

### LINKED ISSUE #255: test_combine_fovs_to_hcs fails

ziw-liu:
Fix #255.

The original test fails due to an ome-zarr-py bug: https://github.com/ome/ome-zarr-py/issues/248#issuecomment-2738712765

### DIFF
--- a/tests/ngff/test_ngff.py
+++ b/tests/ngff/test_ngff.py
@@ -12,7 +12,7 @@ import hypothesis.strategies as st
 import pytest
 import zarr
 from hypothesis import HealthCheck, assume, given, settings
-from numpy.testing import assert_array_almost_equal
+from numpy.testing import assert_array_almost_equal, assert_array_equal
 from numpy.typing import NDArray
 from ome_zarr.io import parse_url
 from ome_zarr.reader import Reader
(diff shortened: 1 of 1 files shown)


ieivanov:
Thanks! This test no longer uses `ome_zarr.reader` - is that OK? I think that was the only test that used that external reader, which I thought was a way to ensure that iohub zarr stores can be read by the "standard" reader. Do we have (or need) and independent test for that?

ziw-liu:
Added a separate test that doesn't trigger the bug.

ieivanov (approved):
Thank you!

PR #285, opened 2025-03-24
Set default pixel size for old NDTiff format

ziw-liu:
Use a default (1.0 micrometer) for NDTiff datasets generated by old MM builds. This will be wrong for most datasets but at least rescues the pixels.

### DIFF
--- a/iohub/ndtiff.py
+++ b/iohub/ndtiff.py
@@ -100,7 +100,9 @@ class NDTiffDataset(MicroManagerFOVMapping):
             )
         self.stage_positions = self._mm_meta["Summary"]["StagePositions"]
         z_step_size = float(self._mm_meta["Summary"]["z-step_um"] or 1.0)
-        xy_pixel_size = float(self._mm_meta["Summary"]["PixelSize_um"] or 1.0)
+        xy_pixel_size = float(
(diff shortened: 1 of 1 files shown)


talonchandler (approved):
Works! Thank you for the quick fix @ziw-liu. 

To make sure I understand this change: dict["key"] returns `KeyError` when "key" doesn't exist, and dict.get("key") returns `None`?

ziw-liu:
> To make sure I understand this change: dict["key"] returns KeyError when "key" doesn't exist, and dict.get("key") returns None?

That's right.

iohub/ndtiff.py
```diff
@@ -100,7 +100,9 @@ def __init__(self, data_path: Path | str):
             )
         self.stage_positions = self._mm_meta["Summary"]["StagePositions"]
         z_step_size = float(self._mm_meta["Summary"]["z-step_um"] or 1.0)
```
ieivanov, line 102:
```suggestion
        z_step_size = float(
            self._mm_meta["Summary"].get("z-step_um") or 1.0
        )
```

ziw-liu: It's already `None` for these files:

```json
        "z-step_um": null
```

PR #287, opened 2025-03-29
Set scale for any coordinate transform

### LINKED ISSUE #278: Set-scale is hard-coded to the highest resolution
#228 introduced a bug, where the 'existing transform' is always taken from the first level (`datasets[0]`):

https://github.com/czbiohub-sf/iohub/blob/354ee0acd6f6524b2f2ca2e1b24d0b3e01324b7c/iohub/ngff/nodes.py#L1167-L1170

This will be wrong when the `image` is higher up in the pyramid.

ziw-liu:
Fix #278.

`Position.set_scale` now works for any level, like `Position.set_transform`.

### DIFF
--- a/iohub/cli/cli.py
+++ b/iohub/cli/cli.py
@@ -123,12 +123,14 @@ def convert(input, output, grid_layout, chunks):
     type=float,
     help="New x scale",
 )
+@click.option("--image", required=False, help="Image name to set scale for")
 def set_scale(
     input_position_dirpaths,
     t_scale=None,
     z_scale=None,
     y_scale=None,
     x_scale=None,
+    image=None,
 ):
     """Update scale metadata in OME-Zarr datasets.
 
@@ -138,6 +140,8 @@ def set_scale(
 
     >> iohub set-scale -i input.zarr/*/*/* -z 2.0
     """
+    if image is None:
+        image = "0"
     for input_position_dirpath in input_position_dirpaths:
         with open_ome_zarr(
             input_position_dirpath, layout="fov", mode="r+"
@@ -147,12 +151,7 @@ def set_scale(
             ):
                 if value is None:
                     continue
-                old_value = dataset.scale[dataset.get_axis_index(name)]
-                print(
-                    f"Updating {input_position_dirpath} {name} scale from "
-                    f"{old_value} to {value}."
-                )
-                dataset.set_scale("0", name, value)
+                dataset.set_scale(image, name, value)
 
 
 @cli.command(name="rename-wells")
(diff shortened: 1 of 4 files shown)


ieivanov (approved):
This looks good to me, it's best if @talonchandler also takes a look before merging.

ziw-liu:
Also @edyoshikun you asked for the organelle box dataset.

iohub/ngff/nodes.py
```diff
@@ -1144,57 +1145,71 @@ def set_transform(
...
-            self.metadata.multiscales[0].datasets[0].coordinate_transformations
+        if "previous_transforms" not in iohub_dict:
+            iohub_dict["previous_transforms"] = []
+        iohub_dict["previous_transforms"].append(
+            {
+                "image": image,
+                "transforms": [
+                    t.model_dump(**TO_DICT_SETTINGS) for t in transforms
+                ],
+                "modified": datetime.now().isoformat(),
```
edyoshikun, line 1192: Is it worth keeping all of the previous transforms or just the most recent one? Flexible on this.

iohub/ngff/nodes.py
```diff
@@ -1144,57 +1145,71 @@ def set_transform(
...
+            }
         )
+        self.zattrs["iohub"] = iohub_dict
         # Replace default identity transform with scale
-        if len(transforms) == 1 and transforms[0].type == "identity":
+        if transforms == [TransformationMeta(type="identity")]:
```
edyoshikun, line 1197: thank you! this was breaking the viewer, but this fixes it

iohub/ngff/nodes.py
ziw-liu: I made it so that when the CLI modifies multiple axes in a loop it doesn't overwrite itself over and over again.

edyoshikun (approved):
The comments above are not a dealbreaker. 

LGTM! Tested the CLI and the script after bumping my environment to 3.11

edyoshikun:
@ziw-liu I just saw your comments. this all LGTM! thanks

iohub/cli/cli.py
```diff
@@ -123,12 +123,14 @@ def convert(input, output, grid_layout, chunks):
     type=float,
     help="New x scale",
 )
+@click.option("--image", required=False, help="Image name to set scale for")
```
talonchandler, line 126:
```suggestion
@click.option("--image", required=False, help="Image name to set scale for. Default = 0")
```

talonchandler (approved):
Thanks for generalizing this @ziw-liu. I tested on a pyramid, and all is working as expected. 

My only non-blocking suggestion is to consider renaming `image` to `multiscale-level` or similar. My understanding of the [OME-Zarr spec](https://ngff.openmicroscopy.org/0.5/#image-layout) is that they refer to the whole pyramid as an `image`, and they refer to each level of the pyramid as a `multiscale-level`. IMO this renaming would also make it easier to understand what this option is for.

PR #289, opened 2025-04-01
Remove fetching step in build docs action

ziw-liu:
Remove the fetching step since the action already runs run the main branch.


JoOkuma (approved):
LGTM!

PR #293, opened 2025-04-03
Read MMStack with Zarr v3

### LINKED ISSUE #292: Reimplement OME-TIFF reading
Work to move to zarr-python 3 in #275 is now blocked by https://github.com/cgohlke/tifffile/issues/282.

Currently the only mechanism to lazily read ND arrays through tifffile is it's ZarrTiffStore, which is incompatible with zarr-python v3, and there is no plan upstream to add support. We may have to revert the OME-TIFF implementation to parsing the MM index map and doing memory-mapped reading ourselves.

ziw-liu:
Fix #292 by wrapping the `ZarrTiffStore`'s storage mapping with `fsspec` so it can be used with zarr-python 3.

Also fixed deprecated TIFF and Zarr readers.

### DIFF
--- a/iohub/_deprecated/singlepagetiff.py
+++ b/iohub/_deprecated/singlepagetiff.py
@@ -3,6 +3,7 @@ import glob
 import json
 import logging
 import os
+from pathlib import Path
 
 import natsort
 import numpy as np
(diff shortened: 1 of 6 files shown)


ziw-liu:
The converter is still broken.

ziw-liu:
The converter also works now, merging.

PR #295, opened 2025-04-15
Support reading OME-Zarr 0.5

ziw-liu:
- [x] `iohub info` works on single-FOV acquire-zarr output
- [x] add tests for reading validated OME-Zarr v0.5 datasets

Next step (for another PR): support writing OME-Zarr v0.5 datasets.


iohub/ngff/models.py
```diff
@@ -219,7 +219,9 @@ class VersionMeta(MetaBase):
     """OME-NGFF spec version. Default is the current version (0.4)."""
 
     # SHOULD
-    version: Literal["0.1", "0.2", "0.3", "0.4"] = "0.4"
+    version: Literal["0.1", "0.2", "0.3", "0.4", "0.5"] | None = Field(
```
ziw-liu, line 222: Workaround for https://github.com/ome/ngff/issues/309.

mattersoflight:
@ziw-liu, please add tests using @aliddell's acquire-zarr example. Is there an example dataset with OME-NGFF 0.5 metadata from OME?

ziw-liu:
> @ziw-liu, please add tests using @aliddell's acquire-zarr example. Is there an example dataset with OME-NGFF 0.5 metadata from OME?

OME does not provide full datasets that are small. I will need to make one and validate it.

ziw-liu:
@aliddell recent zarr-python and numcodec updates broke many things – they shipped multiple breaking changes in patch versions! Also in the news, all zarr-python versions in [3.0.0, 3.0.7] got yanked from PyPI so I'm bumping the requirement.

ziw-liu:
@aliddell is it easy(-ish) to write a test case using acquire-zarr's python bindings to write and using iohub to read?

ziw-liu:
After manually adding metadata omitted by https://github.com/ome/ome2024-ngff-challenge/issues/95. This branch can be used with napari-iohub to visualize v0.5 HCS plates:

```
/hpc/reference/imaging/idr/75-46.ome.zarr
```

image

tests/conftest.py
```diff
@@ -135,3 +137,108 @@ def csv_data_file_2(tmpdir):
...
+    stream = aqz.ZarrStream(settings)
+    data = np.random.randint(
+        0, 2**16 - 1, (32, 4, 10, 48, 64), dtype=np.uint16
+    )
+    stream.append(data)
+    del stream
+
+    yield store_path
+
+    shutil.rmtree(str(store_path))
```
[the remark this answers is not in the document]
aliddell: Good point.

.github/workflows/pr.yml
```diff
@@ -6,7 +6,7 @@ name: lint, style, and tests
 on:
   pull_request:
     branches:
-      - main
+      - "*"
```
aliddell, line 9: Do we really want PRs to all branches to trigger tests? Wouldn't it be better to have a naming convention for development branches?

iohub/ngff/nodes.py
```diff
@@ -1930,14 +1937,56 @@ def rename_well(self, old: str, new: str):
...
+    parse_meta = False
+    if mode in ("r", "r+"):
+        parse_meta = True
+    elif mode == "w-":
+        if store_path.exists():
+            raise FileExistsError(store_path)
+    elif mode == "w":
+        if store_path.exists():
+            if (
+                ".zarr" not in str(store_path.resolve())
```
aliddell, line 1956: Does `store_path` contain an output key as well? e.g., would `/foo/bar/baz.zarr/qux` count? Or do we want the path to end with `.zarr`?

aliddell (approved):
LGTM I just have a couple of questions and a suggestion you can take or leave.

iohub/ngff/nodes.py
ziw-liu: I think things like `plate.zarr/row/column/fov` should count.

PR #297, opened 2025-05-06
Enable tests with ome-zarr-py

ziw-liu:
Tests passes locally with https://github.com/ome/ome-zarr-py/pull/404.

Also the `Plate.from_positions` method is actually broken.


srivarra (approved):
Looks good to me.

PR #298, opened 2025-05-07
Catch json decoding error from incomplete NDTIFF

ziw-liu:
Allows conversion of:

```
/hpc/instruments/jacobo.isim/Notch/20250501_cldnb-lyn-mScar_myo6b-bactin-gfp_she-h2b-gfp_Notch_LY40nM_timelapse/50_hpf_1
```

### DIFF
--- a/iohub/ndtiff.py
+++ b/iohub/ndtiff.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import warnings
+from json import JSONDecodeError
 from pathlib import Path
 from typing import Any, Iterable, Literal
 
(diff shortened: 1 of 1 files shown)


akila-14 (commented):
✅ Looks good to me!
Very rarely, but iSIM acquisitions can end unexpectedly. The default iohub behavior was to quit on such datasets, so this fix is very helpful. The warning message is clear and informative, and it’s great that the rest of the processing can continue even when metadata is incomplete.

Tested on my end with a partially written dataset - works as expected.

Nice and clean fix - thanks, Ziwen!

akila-14 (approved):

PR #302, opened 2025-05-19
Catch truncated unicode error in NDTIFF

ziw-liu:
@akila-14 can you help test this?


akila-14 (approved):
Tested this with a dataset that had incomplete metadata - works well! 👍
Thanks for the fix.

PR #303, opened 2025-05-22
Pin tifffile for zarr-python 2 compatibility

ziw-liu:
This should be applied to the current main (depending on zarr-python 2) and shipped in v0.2.1 to fix installations:

https://github.com/cgohlke/tifffile/issues/297#issuecomment-2899801799


aliddell (approved):
lgtm

Issue #304, opened 2025-05-27
capture the time interval between frames and units (min, second) in ome-zarr metadata


mattersoflight:
At the discussion with Eitan/Assaf, we realized that we should capture the approximate time resolution and units for time-series acquisition.

PR #308, opened 2025-06-03
Convert time scale for OME-TIFF

ziw-liu:
NDTiff does not seem to carry the same information. The unit is always in seconds to prevent mismatch in automated analysis.

Related: #103, #304.

### DIFF
--- a/iohub/convert.py
+++ b/iohub/convert.py
@@ -263,9 +263,11 @@ class TIFFConverter:
         return tuple(chunks)
 
     def _scale_voxels(self):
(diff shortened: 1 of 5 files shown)


ieivanov (approved):
In [359a868](https://github.com/czbiohub-sf/iohub/pull/308/commits/359a8687be3a2b9f483cafdb7f8c1a81bff19680) I moved around the t_scale property to make the definition consistent with zyx_scale and avoid overwriting inherited properties where it's not needed. I think this PR looks good.

PR #310, opened 2025-06-25
Write OME-Zarr 0.5

ziw-liu:


ziw-liu:
cc @bentaculum

aliddell (approved):
A question for you, but looks good.

ziw-liu:
@talonchandler @JoOkuma Here's a daxi2 volume converted so that it has 90x smaller chunks but 10x fewer files:

```
/hpc/projects/comp.micro/test-iohub/daxi/laser_561.zarr
```

mattersoflight:
@ziw-liu thanks for your careful work with this. 
If I remember @aliddell's benchmark, acquire-zarr writes sharded zarr arrays faster than tensorstore, but doesn't require many dependencies for build. 

@aliddell Is acquire-zarr API a drop in replacement for parts of the code that execute the file i/o. I haven't had a chance to read the code closely, but I am curious if the zarr layout generation and metadata creation are sufficiently separated from file i/o that we can improve the performance for sharded arrays using acquire-zarr.

ziw-liu:
Merging into the staging branch (#301). We will handle performance in a future PR.

PR #315, opened 2025-07-12
More robust conversion

ieivanov:
Conversion will continue even if:
 * Data at the first position is missing
 * Data for all positions at a given timepoint is missing

### DIFF
--- a/iohub/convert.py
+++ b/iohub/convert.py
@@ -342,13 +342,7 @@ class TIFFConverter:
                 )
             missing_data_warning_issued = False
             for z_idx in range(self.z):
-                metadata = None
-                try:
-                    metadata = fov.frame_metadata(t=t_idx, c=c_key, z=z_idx)
-                except ValueError:
-                    # This T/C/Z index does not exist,
-                    # warning will be issued below
-                    pass
+                metadata = fov.frame_metadata(t=t_idx, c=c_key, z=z_idx)
                 if metadata is None:
                     if not missing_data_warning_issued:
                         missing_data_warning_issued = True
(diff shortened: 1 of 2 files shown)


iohub/convert.py
```diff
@@ -342,7 +342,13 @@ def _convert_image_plane_metadata(self, fov, zarr_name: str):
                 )
             missing_data_warning_issued = False
             for z_idx in range(self.z):
-                metadata = fov.frame_metadata(t=t_idx, c=c_key, z=z_idx)
+                metadata = None
+                try:
```
ziw-liu, line 346: If this is only a workaround for NDTIFF, maybe catch the error in the reader?

ziw-liu: It's not clear to me what is throwing `ValueError` here.

ieivanov: It is somewhat specific to NDTiff datasets. The value error comes from `_check_coordinates`, which is designed to issue a warning if you request t=0 when there is no time index, but throw an error if you request t>0. In this case, I had a dataset with a timelapse, but all positions at a given timepoint (t=5) were missing, so t_idx=5 was not a valid coordinate. I agree that this try/catch is somewhat a shortcut, the proper way would be to iterate through `self._ndtiff_axes['time']` rather than thru `range(self.t)` but that seems much uglier if this error only happens in edge cases. Let me know if you have other suggestions.

ieivanov: Ah, that should be handled within `NDTiffFOV.frame_metadata`, I'll fix it

iohub/ndtiff.py
```diff
@@ -316,7 +319,7 @@ def xdata(self) -> XDataset:
 
     def get_image_metadata(
         self, p: int | str, t: int, c: int | str, z: int
-    ) -> dict:
+    ) -> dict | None:
```
ziw-liu, line 322: Also need to update the docstring.

ziw-liu (approved):

PR #326, opened 2025-08-22
Fix tensorstore empty array handling

ieivanov:
Fixing a bug where writing writing with `_save_transformed` throws an error when all chunks returned by `_apply_transform_to_czyx` are `None`


ziw-liu:
```
Co-Authored-By: Claude <noreply@anthropic.com>
```

Not sure if we want to include this in the git tree.

ziw-liu (approved):
Thanks for catching this edge case! Please remove the LLM authorship from the commit message before merging.

PR #332, opened 2025-09-10
Pass kwargs to tensorstore

ziw-liu:
Replace the concurrency argument with kwargs for `tensorstore.open`. This provides more flexibility, e.g. for https://github.com/mehta-lab/VisCy/pull/282.

This is a breaking change (compared to 0.3.0a1) but I don't think anyone is directly using it outside of iohub.


srivarra (approved):
LGTM

Issue #345, opened 2025-10-27
Unpickling ImageArray raises exception in v0.2.2 release


applio:
In the tagged v0.2.2 release, when an `ImageArray` is successfully pickled and then later an attempt is made to unpickle it, a `TypeError` exception is raised.  This limits the usability of dependent packages such as [VisCy](https://github.com/mehta-lab/VisCy) which employs the Python Standard Library module `multiprocessing` for parallel processing tasks which in turn pickles and unpickles `ImageArray` objects.

Expected Behavior:
When an `ImageArray` object or any other Python object can be successfully pickled, it should be possible to subsequently unpickle it without triggering an exception.

Observed Behavior:
Using the v0.2.2 release of iohub, a pickled `ImageArray` object can not be successfully unpickled afterwards.

To Reproduce:
Create an `ImageArray` instance, pickle it, then attempt to unpickle it such as in the following example:
```
import iohub, zarr, pickle
za = zarr.array([1, 2, 3, 4])
ia = iohub.ngff.nodes.ImageArray(za)
serialized_ia = pickle.dumps(ia)
# Next line will trigger:  TypeError: ImageArray.__init__() got an unexpected keyword argument 'store'
pickle.loads(serialized_ia)
```

More Details on Impact:
* In particular, VisCy is limited to use with the "fork" start method in `multiprocessing` which has historically been the default on Linux systems but is not available on Windows and has not been the default start method on MacOS systems for some time; when "fork" is used with `multiprocessing`, VisCy does not unpickle `ImageArray` objects at all thanks to the nature of how forking of processes works
* VisCy project issue documenting users encountering the above described issue:  https://github.com/mehta-lab/VisCy/issues/237
* iohub project issue documenting this same behavior which was worked around for a specific use case for Dask but did not address the source of the issue:  https://github.com/czbiohub-sf/iohub/pull/261
* It is believed that the code under active development in iohub's main branch does not reproduce this issue but it is not yet generally available for use in packages like VisCy

PR #346, opened 2025-10-27
Fix pickle support in ImageArray in tag v0.2.2 and add read-only URL support

applio:
Provides a fix to issue #345.  The described issue occurs because `ImageArray` inherits `__setstate__` from its parent but defines its own `__init__` with an incompatible calling signature.  This minimally invasive patch adds flexibility to the calling signature of `ImageArray.__init__` to accomodate the expectations of the inherited `__setstate__` when it triggers the calling of the `__init__`.

In our testing, this patch permits successful unpickling of pickled `ImageArray` instances and enables the current release of the [VisCy](https://github.com/mehta-lab/VisCy) package to support `multiprocessing` start methods other than "fork", allowing VisCy to be successfully used on MacOS and Windows as well as Linux systems.

This PR fixes an issue in the v0.2.2 release of iohub which is believed to be in active use at least until the current work in the main branch stabilizes into a new release with Zarr v3 support.  The in-development code in the main branch has already re-worked the `ImageArray` class to inherit its `__init__` method from its parent, fixing this issue in its own way but because the issue in the current v0.2.2 release is impacting users (see issue #345) it might be worth considering this patch and an update to the v0.2.2 release in the meantime.  As this issue is specific to tag v0.2.2 but there is no appropriate branch to target with this PR, it is currently targeting the main branch but would be much more appropriate if pointed at a branch starting at the v0.2.2 release tag.  Apologies for the confusion.

This solution was originally proposed by @ieivanov in PR #261 but that PR was unfortunately abandoned when a specific use case involving Dask was addressed in PR #262.  This PR merely revives what @ieivanov had originally proposed and credit should go to @ieivanov for working out and addressing the root cause of the problem.

### DIFF
--- a/iohub/ngff/nodes.py
+++ b/iohub/ngff/nodes.py
@@ -60,10 +60,6 @@ def _open_store(
     version: Literal["0.1", "0.4"],
     synchronizer=None,
 ):
-    if not os.path.isdir(store_path) and mode in ("r", "r+"):
-        raise FileNotFoundError(
-            f"Dataset directory not found at {store_path}."
-        )
     if version != "0.4":
         _logger.warning(
             "IOHub is only tested against OME-NGFF v0.4. "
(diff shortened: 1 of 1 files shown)


srivarra:
@applio 

One adjustment:

I will cherry-pick the additional commit (7dafd2d - Read-only URLs) from https://github.com/applio/iohub/tree/support_url_readonly_with_non_fork on top of this one, then retarget the PR to release/v0.2.3.

PR #348, opened 2025-10-28
Added `Position.compute_pyramids`

srivarra:
Adds `Position.compute_pyramids`

### DIFF
--- a/iohub/_deprecated/upti.py
+++ b/iohub/_deprecated/upti.py
@@ -28,7 +28,7 @@ class UPTIReader(ReaderBase):
             self.files = glob.glob(os.path.join(folder, "*.tif"))
         info_img = tiff.imread(self.files[0])
         self.dtype = info_img.dtype
-        (self.height, self.width) = info_img.shape
+        self.height, self.width = info_img.shape
         self.positions = 1
         self.frames = 1
         self.patterns = 0
--- a/iohub/ngff/utils.py
+++ b/iohub/ngff/utils.py
@@ -591,11 +591,9 @@ def process_single_position(
     # Check for invalid times
     time_ubound = input_data_shape[0] - 1
     if np.max(input_time_indices) > time_ubound:
-        raise ValueError(
-            f"""input_time_indices = {input_time_indices} includes
+        raise ValueError(f"""input_time_indices = {input_time_indices} includes
             a time index beyond the maximum index of
-            the dataset = {time_ubound}"""
-        )
+            the dataset = {time_ubound}""")
 
     # Write extra metadata to the output store
     extra_metadata = kwargs.pop("extra_metadata", None)
@@ -755,10 +753,8 @@ def _downsample_tensorstore(
     try:
         import tensorstore as ts
     except ImportError:
-        raise ImportError(
-            "Tensorstore is required for downsampling. \
-            Please install it with `pip install tensorstore`."
-        )
+        raise ImportError("Tensorstore is required for downsampling. \
+            Please install it with `pip install tensorstore`.")
 
     downsampled = ts.downsample(
         source_ts, downsample_factors=downsample_factors, method=method


ieivanov (changes requested):

pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
     "blosc2",
     "xarray>=2024.1.1",
     "dask[array]",
+    "tensorstore>=0.1.64",
```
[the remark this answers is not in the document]
srivarra: That's correct, sure I can do that.

iohub/ngff/nodes.py
```diff
@@ -200,7 +200,7 @@ def __delitem__(self, key):
         """.. Warning: this does NOT clean up metadata!"""
         key = normalize_path(str(key))
         if key in self._member_names:
-            del self[key]
+            del self._group[key]
```
ieivanov: Could you explain why this change is needed?

iohub/convert.py
```diff
@@ -248,6 +248,30 @@ def _gen_chunks(self, input_chunks):
                 f"Chunk type {type(input_chunks)} is not supported."
             )
 
+        # Validate and adjust chunks to be compatible with Dask as they
```
ieivanov: I think these are good checks but may be misplaced. Chunks are again reshared below to ensure that they are less than MAX_CHUNK_SIZE. I think these check should happen after? What is the dim size is odd or prime number? How would you adjust chunk size then to meet Dask requirements? Please add tests that this function executes correctly.

tests/test_converter.py
```diff
@@ -60,7 +60,20 @@ def _check_chunks(
         case "XYZ" | None:
             assert img.chunks == (1,) * 2 + img.shape[-3:]
         case tuple():
-            assert img.chunks == chunks
+            # Chunks may be adjusted to ensure compatibility with data dimensions
```
ieivanov: Please refactor such that you don't have to repeat code during testing.

ieivanov:
Sorry if I'm being a pain in the neck during this review, I think `iohub` follows higher standards than some of our other code. I hope my requests make sense.

srivarra:
> Sorry if I'm being a pain in the neck during this review, I think iohub follows higher standards than some of our other code. I hope my requests make sense.

No I prefer this haha, I'm learning a lot from you so I think it's a good thing tbh.

iohub/ngff/nodes.py
ieivanov: Awesome, thanks for explaining!

iohub/ngff/nodes.py
```diff
@@ -1078,6 +1078,91 @@ def initialize_pyramid(self, levels: int) -> None:
...
+            if num_arrays == 1:
+                raise ValueError(
+                    "Pyramid structure doesn't exist and levels=None. "
+                    "Specify 'levels' parameter to create pyramid."
+                )
+            levels = num_arrays
+
+        if num_arrays == 1:
+            self._initialize_pyramid(levels=levels)
+        elif num_arrays != levels:
```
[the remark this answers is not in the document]
srivarra: If it already has pyramids, should we just throw an error to the user?

ieivanov:
I think we are very close! I'm not fully done reviewing the PR, but before I forget - it looks like `_delete_pyramids` deletes the zarr entries, but does not modify the ome metadata. Please extend `_delete_pyramids` such that it leaves a proper ome zarr store, even if you don't decide to recreate the pyramid levels

ieivanov:
I think this PR looks good! Sorry, I'll change my mind on one thing - we should rename `_initialize_pyramid` back to `initialize_pyramid`. This way we are not introducing a breaking API change in this PR. Plus, the method may be useful if folks want to use iohub to initialize the pyramid structure and create ome metadata, but then compute them some other way.

Before we merge I'd love to get a review from @JoOkuma. 

We should also refactor https://github.com/czbiohub-sf/biahub/pull/12 to use these latest changes - this will tell us if we've gotten the API right.

ieivanov:
I wanted to resurface this PR, @JoOkuma we'd like to get a review from you if you can take a quick look.

PR #351, opened 2025-11-18
Draft convert RGB images

ieivanov:
This is a draft / hack PR where I map Samples in RGB imaged to Channels. I think it may be possible to acquire two channels, where one channel is a standard grayscale fluorescence and the other channels is an RGB image, in which case this workaround would break. I've also commented out the image plane metadata conversion which currently doesn't work. Testing with data in `/hpc/instruments/cm.hummingbird/Amanda/2025_11_06_HandE_stitchng_test/HandE_stitch_test_3/`, converting in `/hpc/projects/comp.micro/exploratory/2025_11_06_HandE_stitchng_test/`


srivarra:
@ieivanov Is this feature still useful for you?

PR #358, opened 2025-12-17
Fix integer types during conversion

ieivanov:
New versions of the zarr library are sensitive to the integer types in chunks - python `int` vs `np.int`, in this case we had a mismatch within the chunks tuple which throws an error


srivarra (approved):
I'm surprised this could even cause a problem.

Issue #265, opened 2024-12-11
Nice to have: `__repr__` functions for ngff nodes


ieivanov:
The `iohub` MicroManager FOV classes have a `__repr__` function which prints information about the dataset. It would be nice to add that functionality for the ngff nodes as well, printing similar information to the `iohub info` CLI call. This can allow us to refactor the `reader.print_info` function.

PR #359, opened 2025-12-23
Opening HCS Plates from a Remote Store

### LINKED ISSUE #209: support ome-ngff on s3
It would be great if this library could access ome-ngff data stored on AWS S3, google GCS, etc. Since you have `fsspec` in your dependencies, I think this would be doable without drastic changes.

- [this logic](https://github.com/czbiohub-sf/iohub/blob/a4d43a2a8cd8d4abce00e475f9f1699d451eba9f/iohub/ngff.py#L51) uses `os.path`, so you would need to use something that's agnostic to the storage backend. But it's not clear why you even need to check if the path exists -- Zarr will do that for you, once you choose a `store` class.
- you are using [DirectoryStore](https://github.com/czbiohub-sf/iohub/blob/a4d43a2a8cd8d4abce00e475f9f1699d451eba9f/iohub/ngff.py#L66). There are two issues with this: first, `DirectoryStore` is not recommended for anything because it puts all the chunks in a single directory and, with too many chunks, that directory becomes unwieldy. Consider using [`NestedDirectoryStore`](https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.NestedDirectoryStore) as a replacement. Second, if you have `fsspec` installed along with zarr, then zarr can transparently store stuff on cloud backends with [`FSStore`](https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.FSStore). `FSStore` will even automatically parse a string like `s3://bucket-name/foo.zarr` and pick the correct backend (s3, in this case).

If there's appetite, I could submit a PR (I didn't see a pre-existing one, but maybe I didn't look too closely).

srivarra:
Added *read-only* support for opening OME-Zarr datasets from remote URLs (HTTP/HTTPS, S3, GCS, Azure).

In addition I made a small adjustment for to use metadata-based navigation instead of directory iteration (HTTP directory listings return HTML artifacts, so we read paths from OME-NGFF metadata files instead)

Here's how you use it:

```python

# Open from public dataset
dataset = open_ome_zarr(
    "https://public.czbiohub.org/comp.micro/neurips_demos/waveorder/20x.zarr/",
    mode="r"
)
```

or with the cli:

```shell
iohub info https://public.czbiohub.org/comp.micro/neurips_demos/waveorder/20x.zarr/ -v
```

I touched up the CLI output a bit as well since it used directory based traversal instead of metadata based one.
- Tree hierarchy now uses rich.Tree instead of manual box-drawing characters
- Summary displayed in a styled Panel with a two-column Table
- Abbreviated Space units (`µm`, `nm`, etc.) now read from OME-Zarr metadata instead of hardcoded to `um`.

```shell
Before:
=== Summary ===
Format:                  omezarr v0.4
(Z, Y, X) scale (um):    (0.1625, 0.1625)
...

After:
╭─────────────── Summary ───────────────╮
│  Format              omezarr v0.4    │
│  (Z, Y, X) scale (µm)  (0.1625, ...) │
│  ...                                 │
╰───────────────────────────────────────╯
```

Todos

- [x] Make sure iohub info works for remote stores over HTTP/HTTPS
- [ ] Add tests for S3, GCS, Azure... (maybe?)

Maybe in the future we can try to wrangle how we would want to do writes. It's also a good idea to address #265 after since I've touched a lot of the info printing code here, so it seems like relative modification.

As an aside I'm using `fssspec`, but `obstore` provides [greater performance](https://earthmover.io/blog/i-o-maxing-tensors-in-the-cloud) some of the time, however it doesn't work for simple servers like `caddy`, but it's great for S3 and such.

### DIFF
--- a/iohub/cli/cli.py
+++ b/iohub/cli/cli.py
@@ -6,6 +6,7 @@ from iohub import open_ome_zarr
 from iohub._version import __version__
 from iohub.cli.parsing import input_position_dirpaths
 from iohub.convert import TIFFConverter
+from iohub.ngff.nodes import _is_remote_url
 from iohub.reader import print_info
 from iohub.rename_wells import rename_wells
 
@@ -14,6 +15,24 @@ VERSION = __version__
 _DATASET_PATH = click.Path(
     exists=True, file_okay=False, resolve_path=True, path_type=pathlib.Path
 )
+
+
+class _PathOrURL(click.ParamType):
+    """Accepts local directory paths or remote URLs."""
+
+    name = "PATH_OR_URL"
+
+    def convert(self, value, param, ctx):
+        if value is None:
+            return None
+        if _is_remote_url(value):
+            return value
+        path = pathlib.Path(value).resolve()
+        if not path.exists():
+            self.fail(f"'{value}' does not exist.", param, ctx)
+        if not path.is_dir():
+            self.fail(f"'{value}' is not a directory.", param, ctx)
+        return path
 
 
 @click.group()
@@ -29,7 +48,7 @@ def cli():
     "files",
     nargs=-1,
     required=True,
-    type=_DATASET_PATH,
+    type=_PathOrURL(),
 )
 @click.option(
     "--verbose",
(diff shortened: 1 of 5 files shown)


ieivanov:
@srivarra is this feature still useful? Can you please update the branch and I'll review it?

srivarra:
@ieivanov Being able to read from the web is still useful, but I think I want to go for a different approach instead, going to close this and I think I'll try something else.

PR #361, opened 2026-01-08
Don't let OME t_scale be equal to zero

### LINKED ISSUE #343: iohub convert can set time scale to 0.0 which napari considers invalid
See `/hpc/instruments/cm.mantis/2025_10_20_neuromast_red_light_imaging/epi-bf-450_1/` and `/hpc/instruments/cm.mantis/2025_10_20_neuromast_red_light_imaging/0-convert/epi-bf-450_1.zarr/`. I don't think I'm using the latest iohub version, looking into it

ieivanov:
t_scale equal to zero causes issues with the napari viewer. t_scale of zero is not physically possible, so we default to 1.0 instead.

Fixes #343


srivarra (approved):
Looks like that fixes it, tyty

PR #362, opened 2026-01-15
Add pyramid sharding

ieivanov:
@srivarra I found that the current pyramid code creates zarr v3 pyramid with small chunks and no sharing which results in lots of kb-size files, see `/hpc/projects/intracellular_dashboard/organelle_dynamics/2024_11_21_A549_TOMM20_DENV/2-assemble/zarr-v3/2024_11_21_A549_TOMM20_DENV.zarr/` for example. Here I'm adding sharding options in `initialize_pyramid`, which seems to work. I haven't yet tested what happens if the original store doesn't have sharding, or if the sharding along time decreased the compute efficiency.

### DIFF
--- a/iohub/ngff/nodes.py
+++ b/iohub/ngff/nodes.py
@@ -1067,11 +1067,13 @@ class Position(NGFFNode):
                 _scale_integers(array.chunks, factor), len(shape)
             )
 
-            # TODO: first check if array is sharded?
-            shards = array.shards[:-3] + _scale_integers(
-                array.shards[-3:], factor
-            )
-            shards_ratio = tuple(s // c for c, s in zip(chunks, shards))
+            if array.shards is not None:
+                shards = array.shards[:-3] + _scale_integers(
+                    array.shards[-3:], factor
+                )
+                shards_ratio = tuple(s // c for c, s in zip(chunks, shards))
+            else:
+                shards_ratio = None
 
             transforms = deepcopy(
                 self.metadata.multiscales[0]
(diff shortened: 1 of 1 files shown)


srivarra:
@ieivanov 
> Do you think we need to handle a case where the original array is not sharded

Right now if the original array isn't sharded, then the pyramids generated from it won't be sharded. 

I think we should either:
1. shard the data ourselves for the downstream pyramids,
2. go ahead and create pyramids without sharding and just notify the user.

Your thoughts?

> do you know if there is a way to inspect the tensorstore compute / operations graph, similar to dask compute graphs

TensorStore doesn't have Dask-style task graph inspection since it uses async/eager execution rather than lazy evaluation, so there's no accumulated graph to visualize :( .

I've never used these, but you could some of the following:

1. Try verbose logging by setting

```shell
os.environ['TENSORSTORE_VERBOSE_LOGGING'] = 'file,file_detail'
```

2.(?) It also seems like you can collet metrics with [`ts.experimental_collect_matching_metrics`](https://google.github.io/tensorstore/python/api/tensorstore.experimental_collect_matching_metrics.html)

3. If that doesn't work, then you're best bet is probably something like manual profiling with cProfile or Scalene with SankeViz...

There's also this [cache pool](https://google.github.io/tensorstore/kvstore/zarr3_sharding_indexed/index.html#json-kvstore/zarr3_sharding_indexed.cache_pool) which *may* save you an operation or two per read operation as every read requires an extra read to fetch the shard index (huh til).

PR #366, opened 2026-01-22
fix(iohub/convert.py): preserve original MM position label and use FOV name for omero.name

### LINKED ISSUE #353: Define how omero["name"] should be used
During conversion from Tiff to OME-zarr we rename MicroManager positions to match the HCS row/col/fov standard using `MicroManagerFOVMapping.hcs_position_labels`. This function turns positions named `A1-Site_0` into FOVs at `A/1/0`, for example.

OME-zarr metadata contains an omero["name"] field. We need to define how that field is used. Continuing with the above example, during zarr conversion we write `omero["name"] = A1-Site_0`, even though the FOV name is now `0`, see 

```
/hpc/projects/intracellular_dashboard/viral-sensor/2024_02_13_ZIV_DENV_Timelapse/0-convert/2024_02_13_ZIV_DENV-Timelapse_1.zarr/A/1/0/.zattrs
```

This omero name is not carried through the different reconstruction steps. For example, after phase reconstruction of that dataset `omero["name"]` is set to `0` by `create_empty_plate`.

We should use that field consistently. My vote is to match the HCS FOV name during conversion and set `omero["name"] = 0`, consistent with downstream reconstruction pipelines. Note that the omero metadata is transitional, and may be removed in the future.

cc @talonchandler

### LINKED ISSUE #354: Keep track of original MM position name in zarr metadata
Related to #353

During conversion from Tiff to OME-zarr we rename MicroManager positions to match the HCS row/col/fov standard using MicroManagerFOVMapping.hcs_position_labels. This function turns positions named A1-Site_0 into FOVs at A/1/0, for example.

We need to keep track of the original MM position name to reference other MM metadata, mainly the stage position list used during stitching. @talonchandler and I discussed that we should log the original MM position label in an `iohub` metadata field that's part of the zarr attributes, similar to how we log the "previous" scale during `iohub set-scale`. This new metadata field could be called `mm_position_label`.

srivarra:
Since those two issues are closely related, and touch the same part of the codebase, It's a bit easier to combine the fix together.


ieivanov:
Is this still useful? Let's merge it?

srivarra:
@ieivanov had to go back and figure out what this was even for. Looks like it's for handling conflicting position names during MM to OME-Zarr conversion, and adding MM's `A1-Site_0` as a property for a FOV inorder to stitch back to MM stage coordinates.

> Is this still useful? 

Yeah. Lemme update and merge.

ieivanov:
A agree that's still needed, just re-read the original issues. Thanks for adding the fix

srivarra:
@ieivanov can you give it a quick review please 🙏 then i'll merge.

ieivanov (approved):
Looks good, thanks!

PR #368, opened 2026-01-28
Prevent data loss during YX to ZYX rechunking for conversion

srivarra:
TIFF files are read slice-by-slice (YX chunks), but ZYX chunking requires combining Z slices via dask rechunking. When dask's intermediate chunk sizes don't divide evenly into the target Zarr chunks, data is silently lost. Fixed by setting `array.chunk-size` config to match the target chunk size before rechunking.

### DIFF
--- a/iohub/convert.py
+++ b/iohub/convert.py
@@ -1,5 +1,6 @@
 import json
 import logging
+from importlib.metadata import version as _get_package_version
 from pathlib import Path
 from typing import Literal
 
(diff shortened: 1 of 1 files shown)


srivarra:
@Soorya19Pradeep 

I believe we should update everything to handle Zarr 3 instead. What are your thoughts?

If you really need a Zarr V2 now, I can add a flag in the convert cli to specify an OME-Zarr version:

i.e. `iohub convert -v 0.5 -i ... -o ...`

Soorya19Pradeep:
Thank you @srivarra ! I think we will eventually definitely move to v3. But for the quick iteration now, I would like to move with zarr v2. It will be great if you have that flag for now.

PR #370, opened 2026-02-09
Add xarray integration to Position

### LINKED ISSUE #369: Allow `Position`s to read/write `xarray`
WaveOrder reconstructions consume CZYX arrays, while `iohub.Position` objects are TCZYX arrays, which leads to passing entire `Positions` (when a smaller CZYX would do) and clunky subsetting operations on metadata-free arrays.   

WaveOrder would benefit from the ability to select subsets of timepoints and channels while keeping track of basic metadata (channel names, physical scales, units). `iohub.Position` objects that can read/write `xarray`'s `DataArray` is one solution: `DataArray`s provide labeled dimensions (TCZYX), clean `.sel` operations, physical coordinates derived from OME scales/translations, and per-coordinate attributes for units.

talonchandler:
Fixes #369.

Add `Position.to_xarray()` and `Position.write_xarray()` methods. `to_xarray()` exports data as a dask-backed `xr.DataArray` with TCZYX dims, physical coordinates from OME scales/translations, and CF-convention coordinate units. `write_xarray()` writes a DataArray (or a subset of channels/timepoints) into an existing Position, mapping xarray coordinates back to zarr indices.

**Notes**
- Named `write_xarray` rather than `from_xarray` since `from_*` implies a constructor
- Coordinate units use CF conventions (`xa.coords["Z"].attrs["units"]`), while `xa.attrs` is reserved for value-level metadata (persisted via `iohub_xarray_attrs` zattrs key)

### DIFF
--- a/iohub/ngff/nodes.py
+++ b/iohub/ngff/nodes.py
@@ -1460,11 +1460,8 @@ class Position(NGFFNode):
         t_coords = data_array.coords["t"].values
         t_indices = np.round((t_coords - translation[0]) / scale[0]).astype(int)
 
-        # Write each (T, C) slice into the zarr array
-        img = self[image]
-        for xa_t, zarr_t in enumerate(t_indices):
-            for xa_c, zarr_c in enumerate(c_indices):
-                img[zarr_t, zarr_c] = np_data[xa_t, xa_c]
+        arr = self[image]
+        arr.oindex[t_indices, c_indices] = np_data
 
         # Persist DataArray attrs to zarr for round-tripping
         if data_array.attrs:


iohub/ngff/nodes.py
```diff
@@ -1328,6 +1316,159 @@ def set_contrast_limits(self, channel_name: str, window: WindowDict):
...
+        x_scale = _coord_scale(data_array.coords["X"].values)
+
+        t_trans = float(data_array.coords["T"].values[0])
+        z_trans = float(data_array.coords["Z"].values[0])
+        y_trans = float(data_array.coords["Y"].values[0])
+        x_trans = float(data_array.coords["X"].values[0])
+
+        # Read coordinate units from per-coordinate attrs (CF convention)
+        def _coord_unit(dim, default):
+            return data_array.coords[dim].attrs.get("units", default)
```
srivarra: [nitpick(if-minor)](https://pullpo.io/cc?l=nitpick&d=if-minor) 

We should rip these out and throw them into `ngff/utils`.

iohub/ngff/nodes.py
```diff
@@ -1328,6 +1316,159 @@ def set_contrast_limits(self, channel_name: str, window: WindowDict):
...
+        scale = self.get_effective_scale(image)
+        translation = self.get_effective_translation(image)
+        t_coords = data_array.coords["T"].values
+        t_indices = np.round((t_coords - translation[0]) / scale[0]).astype(int)
+
+        # Write each (T, C) slice into the zarr array
+        img = self[image]
+        for xa_t, zarr_t in enumerate(t_indices):
+            for xa_c, zarr_c in enumerate(c_indices):
+                img[zarr_t, zarr_c] = np_data[xa_t, xa_c]
```
srivarra: [nitpick(non-blocking)](https://pullpo.io/cc?l=nitpick&d=non-blocking) 
The per-slice loop can be replaced with `.oindex`, which has been used a couple of times apparently throughout `iohub` as a write pattern in the codebase ([`utils.py:174`](https://github.com/czbiohub-sf/iohub/blob/d2e41f09545d083d7447c180656d0404f4d7becd/iohub/ngff/utils.py#L174)) and over in ([`nodes.py:969`](https://github.com/czbiohub-sf/iohub/blob/d2e41f09545d083d7447c180656d0404f4d7becd/iohub/ngff/nodes.py#L969)) where `oindex` is a convenience wrapper around [`set_orthogonal_selection`](https://zarr.readthedocs.io/en/stable/api/zarr/array/#zarr.Array.set_orthogonal_selection). 

We can make the adjustment, or not whatever is more readable for you.

srivarra (changes requested):
Looks good overall, I'll do a pass and implement some of the smaller changes! Thanks!

tests/ngff/test_xarray.py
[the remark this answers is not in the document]
talonchandler: All good suggestions. I did my best to reuse hypothesis pieces from `test_ngff`.

talonchandler:
Thanks @srivarra...all very helpful. I think this is ready for re-review.

PR #403, opened 2026-04-22
Implement version-specific default chunk and shard sizes in iohub utils

### LINKED ISSUE #401: Set version-specific default chunk and shard sizes in `create_empty_plate`
For `version=0.4` (zarr v2) we should use:
* chunks size: (1, 1, Z, Y, X), capped in Z to 500 MB as needed. This is out current default

For `version=0.5` (zarr v3) we should use:
* chunk size: (1, 1, 16, 256, 256) - 2 MB in uint16, 4 MB in float32
* shard size: (1, 1, Z, Y, X) during data processing
* shard size: ~(8, 1, Z, Y, X) in assembled stores

These chunk and shard sizes are designed to:
* balance file size (larger files are easier to handle) and IO speeds (smaller chunks are easier to load during visualization and model training)
* allow for parallel data processing - shard size of (1, 1, ...) allows to parallelize across time and channel without needing to write to the same shard from multiple processes
* informed by DCA requirements: https://github.com/chanzuckerberg/dynamic-cell-atlas-specs/blob/main/docs/v0.2/array-standard.rst

v0.5 shard sizes may need to be optimized to fit in the 1 - 5 GB range. The default for `create_empty_plate` should be `shard_size=(1, 1, Z, Y, X)` and we'll overwrite that during concatenation.

@srivarra @talonchandler @edyoshikun @Soorya19Pradeep any comments are welcome

ieivanov:
Closes #401.

## What this changes

Version-specific chunk and shard defaults for `iohub.ngff.utils.create_empty_plate`, plus the test coverage to pin them down.

### `src/iohub/ngff/utils.py`

- **v0.5** (Zarr v3): default chunks = `(1, 1, 16, 256, 256)` clamped to `shape`; default `shards_ratio` yields a shard of `(1, 1, Z, Y, X)` — one shard per `(T, C)` slot, per the DCA spec referenced in #401.
- **v0.4** (Zarr v2): default chunks = `(1, 1, Z, Y, X)` with Z halved until the single-chunk byte size fits under `_V04_MAX_CHUNK_SIZE_BYTES` (500 MB). No sharding, as Zarr v2 has no sharding codec.
- Extracted the branching into `_default_chunks(shape, dtype, version)` and `_default_shards_ratio(shape, chunks)` helpers, and promoted the two tunables to module-level constants `_V05_DEFAULT_ZYX_CHUNKS` and `_V04_MAX_CHUNK_SIZE_BYTES`.
- Removed the `max_chunk_size_bytes` parameter from `create_empty_plate`; the 500 MB cap is now the module-level constant `_V04_MAX_CHUNK_SIZE_BYTES` and applies only to v0.4 chunks (v0.5 chunks are driven by the DCA spec, not a byte cap).
- Updated the docstring to describe the version-specific defaults.

### Breaking change?

Technically yes, but low blast radius. `create_empty_plate(..., max_chunk_size_bytes=...)` is no longer accepted — callers who passed it by keyword will get a `TypeError`. I confirmed that [waveorder](https://github.com/mehta-lab/waveorder) and [biahub](https://github.com/czbiohub-sf/biahub) — the two primary downstream consumers of this function — never set the parameter explicitly, so this change is a no-op for them. No other in-tree callers pass it either (`grep -R max_chunk_size_bytes` in iohub turns up no hits outside `utils.py`).

If we want to be strict about API stability we can keep `max_chunk_size_bytes` as a deprecated passthrough that overrides the constant for v0.4, but given the known downstream callsites don't touch it, removing it is cleaner.

### `tests/ngff/test_ngff_utils.py`

Added explicit tests that pin the defaults so regressions fail deterministically instead of depending on which examples hypothesis happens to draw:

- `test_v05_default_chunks` (parametrized) — asserts the DCA-aligned chunk shape, clamped to each axis.
- `test_v05_default_shards_cover_zyx` — one shard per `(T, C)` for divisible dims.
- `test_v05_default_shards_with_non_divisible_zyx` — shard still covers `Z/Y/X` when they're not multiples of the chunk size.
- `test_v05_explicit_shards_ratio_is_honored` — caller override wins.
- `test_v04_default_chunks_cover_full_zyx` / `test_v04_default_chunks_capped_by_byte_limit` — full-ZYX chunks; Z halves when the chunk would exceed the byte cap.
- `test_v04_default_has_no_sharding` — `arr.shards is None` on v0.4.
- `test_v04_rejects_explicit_shards_ratio` — raises on v0.4.
- `test_process_single_position_on_sharded_v05_store` — end-to-end round-trip through the default-sharded v0.5 layout.
- `test_apply_transform_to_tczyx_on_multi_time_shard` — direct `oindex` write with `shard_t > 1`. No `xfail`: if the write ever stops succeeding, this fires.

Tightened the hypothesis strategies (`apply_transform_czyx_setup`, `process_single_position_setup`) so integer `time_indices` / `channel_indices` are drawn `unique=True` and sorted via `.map(sorted)`. See the "Why hypothesis needed tightening" section below.

## Why hypothesis needed tightening

Before this change, the existing hypothesis-based tests (`test_apply_transform_to_tczyx_and_save` et al.) passed in CI under the new defaults but **only by luck**: `@settings(max_examples=5)` on each test meant CI drew 5 random examples per run, none of which happened to trip the failure. At `max_examples=200` the same tests failed reliably.

Root cause: the new v0.5 defaults make stores sharded, and the write path goes through `arr.oindex[...]`. iohub's active codec pipeline is `ZarrsCodecPipeline` (zarrs is a required dep), which handles most oindex patterns correctly — but it raises `UnsupportedVIndexingError` and falls back to the default `BatchedCodecPipeline` when an indexer is duplicate or unsorted. The fallback pipeline has an upstream bug in its sharding codec merge step: [zarr-developers/zarr-python#2834](https://github.com/zarr-developers/zarr-python/issues/2834).

Direct repro on a sharded v0.5 store under iohub's default config:

| Indexer pattern | Result |
| --- | --- |
| `oindex[[0, 1], [0]]` — unique, sorted | ✅ OK |
| `oindex[[0, 2], [0]]` — unique, non-contiguous but sorted | ✅ OK |
| `oindex[[0, 0], [0]]` — duplicates | ❌ fallback → zarr#2834 |
| `oindex[[1, 0, 2], [0]]` — unsorted | ❌ fallback → zarr#2834 |

The hypothesis strategy was generating duplicate / unsorted integer lists like `[0, 0]` and `[1, 0, 2]`. Real callers (`biahub`, the hybrid_scheme processing pipelines, etc.) always build sorted unique index lists, so the duplicate/unsorted draws weren't exercising a real use case — they were exercising a zarrs fallback path that wouldn't otherwise fire. Enforcing `unique=True` + `.map(sorted)` at the strategy level matches what callers pass and keeps the tests on the zarrs fast path.

## Verification

Full file:

```
pytest tests/ngff/test_ngff_utils.py
19 passed in 3.49s
```

Stress run (all three hypothesis-based tests, `max_examples=200` each, generate-only so the 600 draws are all fresh):

```
3 passed in 140.54s
```

No flakes, no `xfail`, and the new explicit tests fail loudly if any of the pinned defaults regress.

## Test plan

- [x] `tests/ngff/test_ngff_utils.py` passes at default settings
- [x] Hypothesis-based tests pass at `max_examples=200` (no duplicate/unsorted fallbacks)
- [x] v0.5 sharded write path round-trips correctly via `process_single_position`
- [x] v0.5 multi-time oindex write into a shared shard succeeds
- [x] v0.4 rejects `shards_ratio` and has `shards is None` by default
- [x] Confirmed no downstream callers (`waveorder`, `biahub`) pass `max_chunk_size_bytes`
- [x] Test in biahub processing pipeline on mantis v2

## Follow-up
Should we update and `Position.create_zeros` (and therefore `create_image` which calls `create_zeros`) with better chunk and shard defaults? Currently chunking defaults to (1, 1, Z, Y, X), even for v0.5 and don't use sharding by default for v0.5

Transferred to https://github.com/czbiohub-sf/iohub/issues/409

```python
    def create_zeros(
        self,
        name: str,
        shape: tuple[int, ...],
        dtype: DTypeLike,
        chunks: tuple[int, ...] | None = None,
        shards_ratio: tuple[int, ...] | None = None,
        transform: list[TransformationMeta] | None = None,
        check_shape: bool = True,
    ):
        """Create a new zero-filled image array in the position.
        ...
        """
        if not chunks:
            chunks = self._default_chunks(shape, 3)

        if check_shape:
            self._check_shape(shape)
        arr_handle = self._create_zarr_array(name, shape, dtype, chunks, shards_ratio)
        img_arr = ImageArray.from_handle(arr_handle, self._impl)
        self._create_image_meta(name, transform=transform)
        return img_arr

    @staticmethod
    def _default_chunks(shape, last_data_dims: int):
        chunks = shape[-min(last_data_dims, len(shape)) :]
        return pad_shape(chunks, target=len(shape))
```

### DIFF
--- a/src/iohub/ngff/utils.py
+++ b/src/iohub/ngff/utils.py
@@ -19,6 +19,13 @@ from iohub.ngff import open_ome_zarr
 from iohub.ngff.nodes import TransformationMeta
 
 
+#: Default ZYX chunk size for OME-Zarr v0.5 stores: ~2 MB at uint16 / ~4 MB at float32.
+_V05_DEFAULT_ZYX_CHUNKS: tuple[int, int, int] = (16, 256, 256)
+
+#: Maximum chunk size in bytes used for the v0.4 default-chunk calculation.
+_V04_MAX_CHUNK_SIZE_BYTES: float = 500e6
+
+
 def create_empty_plate(
     store_path: Path,
     position_keys: list[tuple[str, str, str]],
@@ -29,7 +36,6 @@ def create_empty_plate(
     version: Literal["0.4", "0.5"] = "0.5",
     scale: tuple[float, ...] = (1, 1, 1, 1, 1),
     dtype: DTypeLike = np.float32,
-    max_chunk_size_bytes: float = 500e6,
 ) -> None:
     """
     Create a new HCS Plate in OME-Zarr format if the plate does not exist.
@@ -50,22 +56,24 @@ def create_empty_plate(
     shape : tuple[int, ...]
         TCZYX shape of the plate.
     chunks : tuple[int, ...], optional
-        TCZYX chunk size of the plate. If None, the chunk size is calculated
-        based on the shape to be less than max_chunk_size_bytes.
+        TCZYX chunk size of the plate. If None, a version-specific default
+        is used:
+        - "0.4": ``(1, 1, Z, Y, X)`` capped in Z to 500 MB.
+        - "0.5": ``(1, 1, 16, 256, 256)`` clamped to ``shape``.
         Defaults to None.
     shards_ratio : tuple[int, ...], optional
-        TCZYX shards ratio of the plate.
-        If None, no sharding is applied.
+        TCZYX shards ratio of the plate (shard size = chunks * shards_ratio).
+        If None, a version-specific default is used:
+        - "0.4": no sharding (Zarr v2 does not support it).
+        - "0.5": ratio that yields a shard size of ``(1, 1, Z, Y, X)``.
         Defaults to None.
     version : Literal["0.4", "0.5"], optional
         OME-Zarr version to use for the plate.
-        Defaults to "0.4".
+        Defaults to "0.5".
     scale : tuple[float, ...], optional
         TCZYX scale of the plate. Defaults to (1, 1, 1, 1, 1).
     dtype : DTypeLike, optional
         Data type of the plate. Defaults to np.float32.
-    max_chunk_size_bytes : float, optional
-        Maximum chunk size in bytes. Defaults to 500e6 (500MB).
 
     Examples
     --------
@@ -101,20 +109,16 @@ def create_empty_plate(
 
     Notes
     -----
-    - If `chunks` is not provided, the function calculates an appropriate
-    chunk size to keep the chunks under the specified `max_chunk_size_bytes`.
+    - If `chunks` is not provided, a version-specific default is used (see
+    parameter description).
     - The function ensures that positions and channels are appended to an
     existing plate if they are not already present.
     """
-    # Limiting the chunking to 500MB
     if chunks is None:
-        chunk_zyx_shape = _limit_zyx_chunk_size(
-            shape,
-            np.dtype(dtype).itemsize,
-            max_chunk_size_bytes,
-        )
+        chunks = _default_chunks(shape, dtype, version)
 
-        chunks = 2 * (1,) + chunk_zyx_shape
+    if shards_ratio is None and version == "0.5":
+        shards_ratio = _default_shards_ratio(shape, chunks)
 
     # Create plate
     output_plate = open_ome_zarr(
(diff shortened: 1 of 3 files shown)


src/iohub/ngff/utils.py
```diff
@@ -19,6 +19,13 @@
 from iohub.ngff.nodes import TransformationMeta
 
 
+#: Default ZYX chunk size for OME-Zarr v0.5 stores: ~2 MB at uint16 / ~4 MB at float32.
+_V05_DEFAULT_ZYX_CHUNKS: tuple[int, int, int] = (16, 256, 256)
+
+#: Maximum chunk size in bytes used for the v0.4 default-chunk calculation.
+_V04_MAX_CHUNK_SIZE_BYTES: float = 500e6
```
srivarra: Could we move this over to `src/iohub/core/compat.py` ?

tests/ngff/test_ngff_utils.py
```diff
@@ -786,3 +800,255 @@ def test_process_single_position(setup, constant, num_threads):
...
+    out_arr = _open_array(output_store, position_key)
+    assert out_arr.shards == (1, 1, shape[2], shape[3], shape[4])
+
+    with open_ome_zarr(input_store) as in_ds, open_ome_zarr(output_store) as out_ds:
+        in_data = in_ds["/".join(position_key)].data[:]
+        out_data = out_ds["/".join(position_key)].data[:]
+    np.testing.assert_array_almost_equal(out_data, dummy_transform(in_data, constant=2))
+
+
+def test_apply_transform_to_tczyx_on_multi_time_shard(tmp_path):
```
srivarra: [question](https://pullpo.io/cc?l=question) 
I'm not exactly sure what we are testing here, are we testing that the Zarr Python `codec_pipeline` doesn't change? Since `ZarrsCodecPipeline` is the default one, it won't ever use `BatchedCodecPipeline` right?

srivarra (changes requested):
Changes look good overall, just a question about what we are testing.

src/iohub/ngff/utils.py
ieivanov: Done!

tests/ngff/test_ngff_utils.py
ieivanov: I renames this test to use sharding across channels rather than timepoints - that's more common for us. In the new `test_apply_transform_to_tczyx_on_multi_channel_shard` I'm testing that we can process in parallel two channels that come from the same shard and need to also go into one shard in the output store. This functionality is not tested in previous tests. I removed references to the upstream zarr issues and `BatchedCodecPipeline` - that was a red herring.

srivarra: Gotcha, thank you for clarifying!

talonchandler (approved):
All looks good to me. 

Friendly reminder to not delete this branch when you merge (waveorder's main now depends on it), then release, then update waveorder's dependency. Thank you!

srivarra (approved):
Looks good!

edyoshikun (approved):

PR #405, opened 2026-04-22
fix: rename zarr entry-point group for pyproject schema

ieivanov:
## Summary
Renames the `[project.entry-points]` group from `iohub.zarr-implementations` to `iohub.zarr_implementations` and updates the registry to match.

## Why
The [SchemaStore pyproject.json](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/pyproject.json) only allows `entry-points` group names matching `^\w+(\.\w+)*$` (dot-separated segments of word characters; hyphens are not allowed in a segment). Editors and tools that validate `pyproject.toml` against that schema (e.g. Taplo) reported: `Additional properties are not allowed ('iohub.zarr-implementations' was unexpected)`.

## Note
**Breaking change** for any external package that registered under the old group name. Update third-party `[project.entry-points."iohub.zarr-implementations"]` to `iohub.zarr_implementations`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ tensorstore = [
     "tensorstore>=0.1.64",
 ]
 
-[project.entry-points."iohub.zarr-implementations"]
+[project.entry-points."iohub.zarr_implementations"]
 zarr = "iohub.core.implementations.zarr_python:ZarrPythonImplementation"
 tensorstore = "iohub.core.implementations.tensorstore:TensorStoreImplementation"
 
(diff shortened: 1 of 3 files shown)


ieivanov:
@srivarra your call if you want this or now

srivarra (approved):
@ieivanov Tyty, I wonder why it never caused an error earlier.

ieivanov:
I'm not sure, I only caught it because by editor complained about its style, it would be the build still worked file

PR #406, opened 2026-04-22
feat(tensorstore): expose recheck_cached_data on TensorStoreConfig

edyoshikun:
Add the `rechecked_cached_data` and fowrad the kward into _ts_open in open_array, when it's explicitly set.


edyoshikun:
@srivarra can we merge this to bump to 0.3.3 or alpha so we can pin on viscy?

srivarra:
@edyoshikun yeah

srivarra (approved):

PR #410, opened 2026-04-25
refactor: add num_workers/use_threads to process_single_position

ieivanov:
## Summary

- Re-introduce a process-pool path in `process_single_position` for tensor-heavy CPU callers (deskew, register, deconvolve), where threads OOM the slurm cgroup because all concurrent task allocations live in one address space and torch's CPU caching allocator never returns memory.
- Replace the old `mp.Pool` with `concurrent.futures.ProcessPoolExecutor(mp_context=spawn)` so silent worker death (e.g. cgroup OOM-kill) surfaces as `BrokenProcessPool` instead of hanging on `pool.starmap` until walltime.
- New public params: `num_workers` (replaces `num_processes` and `num_threads`) and `use_threads` (default `False`). Old names still accepted with `DeprecationWarning` and forwarded to `num_workers`.
- `num_workers <= 1` keeps the in-process serial loop introduced in #396.

### Why processes again

#396 switched to `ThreadPoolExecutor` because the transforms release the GIL. That's true for I/O-bound callers, but for tensor-heavy CPU work the per-task working set (e.g. deskew is ~7 × one input volume after rearrange + grid_sample + mean) lives in one process under threads, and torch's caching allocator does not give the memory back. On a real run on this cluster:

- 16 spawn workers × 30 GB cgroup → ~470 GB cgroup OOM, one worker SIGKILLed, `mp.Pool.starmap` deadlocks waiting for the dead worker. py-spy confirmed: main blocked at `wait → starmap`, all surviving workers idle on `queue.get`. (This is the failure mode #396 replaced threads to avoid; processes still do it under cgroup OOM if you don't surface the error.)
- 16 thread workers in one process → entire process SIGKILLed at MaxRSS ≈ 500 GB on the same budget, no progress.

`ProcessPoolExecutor` solves both: spawn-isolated address space (memory hygiene), and `BrokenProcessPool` propagation (no silent hangs). The threaded path is preserved as `use_threads=True` for I/O-bound callers like the existing `iohub` write paths.

### Behaviour matrix

| `num_workers` | `use_threads` | Path |
|---|---|---|
| `<= 1` | (any) | in-process serial loop (same as #396) |
| `> 1` | `True` | `ThreadPoolExecutor` |
| `> 1` | `False` | `ProcessPoolExecutor(mp_context=spawn)` (new default) |

### Deprecation

- `num_processes` was already deprecated in #396 (forwarded to `num_threads`). Now forwarded to `num_workers` instead.
- `num_threads` is newly deprecated; forwarded to `num_workers`. Callers that wanted threads should also pass `use_threads=True`.

## Test plan

- [x] CI green (existing `test_process_single_position` covers the threaded path; should also pass with `num_workers + use_threads=True`).
- [x] Add a `use_threads=False` parametrization to the existing test so the spawn path is covered.
- [x] Confirm that calling with `num_processes=N` or `num_threads=N` emits `DeprecationWarning` and forwards correctly.

### DIFF
--- a/src/iohub/ngff/utils.py
+++ b/src/iohub/ngff/utils.py
@@ -166,8 +166,8 @@ def _apply_transform_to_czyx(
         kwargs["input_time_index"] = input_time_index
 
     click.echo(f"Processing t={input_time_index}, c={input_channel_indices}")
-    input_dataset = open_ome_zarr(input_position_path, layout="fov", mode="r")
-    czyx_data = input_dataset.data.oindex[input_time_index, input_channel_indices]
+    with open_ome_zarr(input_position_path, layout="fov", mode="r") as input_dataset:
+        czyx_data = input_dataset.data.oindex[input_time_index, input_channel_indices]
     if not _check_nan_n_zeros(czyx_data):
         return func(czyx_data, **kwargs)
     else:
@@ -473,17 +473,28 @@ def process_single_position(
 
 def _check_nan_n_zeros(input_array) -> bool:
     """Checks if any of the channels are all zeros or nans."""
-    if len(input_array.shape) == 3:
-        if np.all(input_array == 0) or np.all(np.isnan(input_array)):
-            return True
-    elif len(input_array.shape) == 4:
-        num_channels = input_array.shape[0]
-        for c in range(num_channels):
-            zyx_array = input_array[c, :, :, :]
-            if np.all(zyx_array == 0) or np.all(np.isnan(zyx_array)):
-                return True
+    if input_array.ndim == 3:
+        return _zyx_is_all_zero_or_nan(input_array)
+    elif input_array.ndim == 4:
+        return any(_zyx_is_all_zero_or_nan(input_array[c]) for c in range(input_array.shape[0]))
     else:
         raise ValueError("Input array must be 3D or 4D")
+
+
+def _zyx_is_all_zero_or_nan(zyx_array) -> bool:
+    """All-zero or all-NaN test that short-circuits on the first counter-example.
+
+    `np.any(arr)` returns False iff every element is 0/False, and short-circuits
+    in C as soon as it finds a truthy value. The previous `np.all(arr == 0)`
+    materialised a full boolean mask of the input volume before reducing it.
+    """
+    if not np.any(zyx_array):
+        return True  # all zeros
+    # NaN is truthy in numpy bool context, so the explicit NaN check is only
+    # needed when np.any returned True (otherwise the array is all zeros and
+    # would not reach here).
+    if zyx_array.dtype.kind == "f" and np.isnan(zyx_array).all():
+        return True  # all NaN
     return False
 
 
--- a/src/iohub/ngff/utils.py
+++ b/src/iohub/ngff/utils.py
@@ -280,6 +280,21 @@ def _slice_to_list(indices: list[int] | slice) -> list[int]:
     return indices
 
 
+def _available_cpus() -> int:
+    """Return the CPU count the current process is allowed to use.
+
+    Slurm exports ``SLURM_CPUS_PER_TASK`` for tasks that ask for more than
+    one CPU, which reflects the cgroup CPU allocation rather than the
+    host's total CPU count. Honouring it here prevents oversubscribing
+    the cgroup when ``os.cpu_count()`` reports the whole node (e.g. 128)
+    while slurm only granted us a few cores.
+    """
+    slurm_cpus = os.environ.get("SLURM_CPUS_PER_TASK")
+    if slurm_cpus and slurm_cpus.isdigit():
+        return int(slurm_cpus)
+    return os.cpu_count() or 1
+
+
 def process_single_position(
     func: Callable[[NDArray, Any], NDArray],
     input_position_path: Path,
@@ -430,8 +445,7 @@ def process_single_position(
         output_position_path,
         **kwargs,
     )
-    cpu_count = os.cpu_count() or 1
-    num_workers = min(num_workers, len(flat_iterable), cpu_count)
+    num_workers = min(num_workers, len(flat_iterable), _available_cpus())
     if num_workers <= 1:
         click.echo("\nRunning serially in the calling process (num_workers <= 1)")
         for args in flat_iterable:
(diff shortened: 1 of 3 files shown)


ieivanov:
@aofei-liu I'm still testing this PR. Let's figure out why nextflow did not work well with spawned processes, and if there is a way to fix that. I'm leaving `use_threads=False` as an option still, that's currently not wired in `biahub`  - please test what works best for nextflow and we can expose that through a CLI argument or env variable for biahub reconstructions.

aofei-liu (approved):

src/iohub/ngff/utils.py
```diff
@@ -287,8 +303,10 @@ def process_single_position(
     output_channel_indices: list[slice] | list[list[int]] | None = None,
     input_time_indices: list[int] | None = None,
     output_time_indices: list[int] | None = None,
+    num_workers: int = 1,
+    use_threads: bool = False,
     num_processes: int | None = None,
-    num_threads: int = 1,
+    num_threads: int | None = None,
```
edyoshikun: seems a bit dangerous. I am more in favor if you just completely deprecate the num_threads. Otherwise you will run into bugs in the future.

ieivanov: Both `num_processes` and `num_threads` if passed get assigned to `num_workers` which then get routed to sequential execution, `ProcessPoolExecutor`, or `ThreadPoolExecutor` depending on the number of workers (`num_workers=1` gives you sequential execution) and whether you set `use_threads` which is False by default. For now i'd like to keep the option for multithreading until we resolve whatever is happening with nextflow. Eventually I agree that we should remove `num_processes` and `num_threads`. Does that sounds good to you?

aofei-liu:
hi! looks good and yes let's figure out what's the issue. 

the issue with multipool for nextflow that i previously observed was that since nextflow is handling the parallelization, child processes in a pool create resource competition that nextflow is not aware of and cause jobs to oom or hang indefinitely when a child job doesn't get allocated a task and never exits. at the time i assumed that most processes were IO-bound and not GIL-bound (i.e. processing itself is fast compared to time for i/o of the image), so multithreading worked well. it seems it's not the case for deskew or any of the gpu-heavy processes, so we should definitely revisit. it also sounds like using `ProcessPoolExecutor` might fix this - let's test

Also from what I remember from before, on my end the spawning step itself was creating unnecessary overhead that was slowing stuff (but this might have been for flatfield, which is a relatively fast operation). i think the answer could also be to just spawn 1 process/thread per job when parallelizing with nextflow and to create higher parallelization at the submitit level (i.e. spawn a bunch of jobs) as a default, instead of trying to juggle both processes/threads + nextflow parallelization. the techdev pipeline i have right now also forces num processes/threads = 1 i think so it shouldn't be breaking outside of a couple of syntax catches

src/iohub/ngff/utils.py
edyoshikun: so what I find confusing is that you have num_workers and num_processes and then you have num_threads which gets overwrittern but you also have the use_threads parameter. I'll let you handle this in a separate PR later but it seems like you should only have `num_workers`, `use_threads` as parameters. 

Also, when you are running the nextflow depending on how you set it up, you have a limited number of threads per node, so for I/O operations this is good, but for other things, if you do threads, since it's unbounded, it might give you this OOM. I think exposing both the `ThreadPool` or `Multiprocess` is fine here, but it just add more configuration in nextflow.

edyoshikun:
@aofei-liu to answer your GPU issue:

One caveat... if a worker holds the GPU for a long synchronous kernel, GIL-released or not, you won't get linear speedup with threads because the GPU is the serializing resource. So num_workers > 1 on GPU mainly helps when there's host↔device transfer to overlap with compute. Often num_workers=1 + a single well-batched GPU call is faster than 8 thread workers fighting over the same SM (which is why we do this trick on VisCy) Now this is if you have enough VRAM....

edyoshikun:
I prefer if we don't have backwards compatibility and we fix the imports. Ill let you guys fix this in a later PR, but I think it just creates more sources of error if you do patchwork.

aofei-liu:
good with me to not have backwards compat - the wire-nextflow stuff is still in a branch, i can just fix on my end

ieivanov:
Sounds good! I'll remove `num_processes` and `num_threads` as arguments to `process_single_position` and we'll be left with just `num_workers` and `use_threads`. We'll update `biahub` to use these arguments when we upgrade to this version of `iohub`. I'll push this change and then we can merge.

tests/conftest.py
```diff
@@ -13,6 +14,19 @@
...
+# tests using ProcessPoolExecutor (e.g. test_process_single_position with
+# use_threads=False) can unpickle helpers like `tests.ngff.test_ngff_utils.
+# dummy_transform`. pytest's `--import-mode=importlib` only manipulates the
+# parent process's sys.path, not the env that spawn children inherit.
+_REPO_ROOT = Path(__file__).resolve().parent.parent
+os.environ["PYTHONPATH"] = os.pathsep.join(
+    [str(_REPO_ROOT)] + ([os.environ["PYTHONPATH"]] if os.environ.get("PYTHONPATH") else [])
+)
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
```
srivarra: I tried running tests with this included and excluded for python 3.13 and 3.14. The tests ran fine across both without a problem. Did you run into an error with unpickling? If so I can't reproduce it.

srivarra (commented):
What was your experience with and without the changes to the conftest?

tests/conftest.py
ieivanov: Hm, let's see, I'll push a commit reverting these changes. My Claude said he was able to reproduce the failing tests.

ieivanov: Now I remember- the failure only happens for `process_single_position(..., use_threads=False, num_workers >= 2)` - it's possible in your run the hypothesis testing didn't draw this combination. If you run this a few times locally you should be able to see the tests fail, as in [0f86c59](https://github.com/czbiohub-sf/iohub/pull/410/commits/0f86c5987d24203c31456b2e0d4455f68b5a2122)

srivarra (approved):
@ieivanov This looks good. I think we're ready to merge, and I'll go ahead and make a new release

PR #418, opened 2026-06-01
feat: add zarr-python as an optional zarr implementation (default is zarrs-python)

### LINKED ISSUE #416: Zarrs Codec deadlocking with  num_workers>0 iwth Torch.Dataloaders
I added a comment here where we are erroring in the first `HCSDataloader`. This is not a viscy issue, so logging it here first. 
https://github.com/dl-janelia/image_translation/pull/17#issuecomment-4560375173

The deadlock is not zarr-python's asyncio loop. It's zarrs — the Rust codec pipeline.

CC: @alxndrkalinin

srivarra:
Lets users pick the Zarr codec pipeline in `open_ome_zarr`:

- `implementation="zarrs-python"`: `zarrs`,  Rust pipeline, default
- `implementation="zarr-python"`:  `zarr-python`, pure-Python pipeline, opt-in

The choice was previously implicit (`zarrs` auto-selected when installed); it's now explicit by name. Default behavior is unchanged.

Slight Breaking change

`implementation="zarr"` is removed, use `"zarrs-python"` or `"zarr-python"` or `"tensorstore"`.

### DIFF
--- a/docs/examples/torch_dataloader.md
+++ b/docs/examples/torch_dataloader.md
@@ -1,10 +1,10 @@
 # Reading OME-Zarr in a PyTorch DataLoader
 
-iohub auto-selects the [zarrs](https://github.com/zarrs/zarrs-python) (Rust)
-codec pipeline whenever the `zarrs` package is installed. It is fast, but its
-parallel codec runs on a process-global thread pool that is **not fork-safe**.
-PyTorch's `DataLoader` forks worker processes on Linux when `num_workers > 0`,
-and the first chunk decode inside a forked worker deadlocks.
+iohub defaults to the `zarrs-python` implementation, which uses the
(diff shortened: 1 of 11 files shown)


edyoshikun (approved):
tyty LGTM

edyoshikun:
we will prioritize using zarrs on viscy as well if it looks like that is the most performant. for now we will just have the option for zarr-python. thanks

srivarra:
@ieivanov Will changing the implementation from `implementation="zarr"` effect any of your work? I'm assuming that you've left it as the default value and didn't specify the `implementation` directly.

ieivanov:
LGTM. @srivarra it'll be great if you can bundle this with https://github.com/czbiohub-sf/iohub/pull/419 into a new release

ieivanov (approved):

PR #419, opened 2026-06-01
feat: add metadata_sources to create_empty_plate

aofei-liu:
## Summary

- Adds an optional `metadata_sources` parameter to `create_empty_plate` that copies per-position **custom (non-OME) zattrs** (e.g. `extra_metadata`) from one or more source HCS plates to the newly created output plate. OME-owned keys (`ome`, `multiscales`, `omero`, `labels`, `version`) — i.e. coordinate transforms, axis definitions, and label references — are **not** copied.
- Accepts a single `Path`/`str` or a `list` of them. Metadata is copied only for **newly created** positions; pre-existing positions are left untouched. A given zattrs key is copied only if it is not already present on the destination, so **earlier sources take precedence** over later ones.
- The source plate root must exist (`FileNotFoundError` otherwise); each source position is opened with `layout="fov"` so a non-FOV source surfaces a meaningful error, while positions simply absent from a source are skipped gracefully.
- This avoids the need for a separate `copy_position_metadata` call after plate creation — the pattern currently repeated across every `_init_output_plate` in [biahub](https://github.com/czbiohub-sf/biahub) (deskew, flat-field, reconstruct, virtual-stain, tracking).
- Follows up on the discussion in https://github.com/czbiohub-sf/biahub/pull/255#issuecomment-4596523191 where @ieivanov suggested folding this into iohub.

## Test plan

- [x] `test_create_empty_plate_copy_metadata_from` — custom zattrs copied; dest dataset layout, scale, and channel names preserved; label references not copied
- [x] `test_create_empty_plate_copy_metadata_subset_positions` — only matching positions get metadata copied when dest has a subset of source positions
- [x] `test_create_empty_plate_copy_metadata_skips_existing_positions` — pre-existing dest positions left untouched
- [x] `test_create_empty_plate_copy_metadata_missing_source_root` — non-existent source raises `FileNotFoundError`
- [x] `test_create_empty_plate_copy_metadata_position_absent_in_source` — positions missing from the source are created without copied metadata
- [x] `test_create_empty_plate_copy_metadata_multiple_sources` — a list of sources merges disjoint zattrs keys
- [x] `test_create_empty_plate_copy_metadata_earlier_source_wins` — on a duplicate key, the earlier source in the list wins
- [x] Existing `test_create_empty_plate` still passes (no regression)


ieivanov:
We'll reduce the scope of this PR to only copy external metadata, for example as written by biahub

aofei-liu:
@ieivanov changed to only copy custom metadata, ready for review

ieivanov (approved):

ieivanov:
@aofei-liu could you take a look, I moved things around a bit. Most importantly, I'm restricting the metadata copy to new positions only. `create_empty_plate` should be no-op is the target position already exists. I think this is ready to merge, once you confirm.

aofei-liu:
@ieivanov thanks! I took a look, I think it makes sense to not overwrite. should I wait for sri to review too?

src/iohub/ngff/utils.py
```diff
@@ -71,6 +75,13 @@ def create_empty_plate(
         TCZYX scale of the plate. Defaults to (1, 1, 1, 1, 1).
     dtype : DTypeLike, optional
         Data type of the plate. Defaults to np.float32.
+    copy_metadata_from : Path or str, optional
```
edyoshikun: this is an interesting one. I think we make the assumption that the plate we are copying has the exact same layout but for the cases where we concatenate and or grab a subset of FOVs, this might not work(?)
For example if you merge two datasets together or you create a new zarr from a subset.

edyoshikun (commented):
I think in principle this should work when you are copyin in and from the same zarr store layout. If you are merging or assembling a new zarr store, not sure if this works.

aofei-liu:
@edyoshikun Addressed the renamed-position metadata issue you flagged.

**New parameter: `copy_metadata_map`** (mutually exclusive with `copy_metadata_from`)

Accepts a `dict[str, Path]` mapping destination position key strings to source position paths. This handles the case where `get_output_paths(..., ensure_unique_positions=True)` renames duplicate positions (e.g. `A/1/0` → `A/1d0/0`), which breaks string-matching against the source plate.

```python
create_empty_plate(
    store_path=output,
    position_keys=[("A", "1d0", "0"), ("A", "1d1", "0")],
    channel_names=["DAPI"],
    shape=(1, 1, 256, 256, 256),
    copy_metadata_map={
        "A/1d0/0": Path("/plate1.zarr/A/1/0"),
        "A/1d1/0": Path("/plate2.zarr/A/1/0"),
    },
)
```

Three new tests cover the mapping, skip-existing, and mutual-exclusivity behaviors. The biahub `refactor-assembly` branch (`d922c28`) wires this into both `_init_concatenate` and `concatenate` by building the map from the existing `all_data_paths` / `output_position_paths` pairing.

src/iohub/ngff/utils.py
```diff
@@ -139,6 +192,25 @@ def create_empty_plate(
...
+                    src_path = copy_metadata_map.get(position_key_string)
+                else:
+                    src_path = copy_metadata_from / position_key_string
+                src_pos = None
+                if src_path is not None:
+                    try:
+                        src_pos = open_ome_zarr(src_path, mode="r")
+                    except FileNotFoundError:
+                        src_pos = None
+                if src_pos is not None:
```
srivarra: [issue(blocking)](https://pullpo.io/cc?l=issue&d=blocking) 

When `position_keys` and `copy_metadata_map` keys are not kept in sync:

```python
position_keys=[
    ("A", "1d0", "0"),     # from plate1
    ("A", "1d1", "0"),     # from plate2
]

copy_metadata_map={
    "A/1d0/0": Path("/data/plate1.zarr/A/1/000000"),
    # "A/1d1/0": ...  <-- Say I forgot to add this entry here
}
```


The empty array for `A/1d1/0` still gets created and looks valid, but that custom metadata silently doesn't get copied over.

srivarra (changes requested):
Looks great overall, just one small thing to fix.

ieivanov (commented):
>I think in principle this should work when you are copying in and from the same zarr store layout. If you are merging or assembling a new zarr store, not sure if this works.

@edyoshikun good point, it will be most useful if all that metadata ends up in the concatenated zarr store, which is assembled from multiple stores. I think we can make `copy_metadata_from: list[Path]` (you'd pass `\path\to\dataset.zarr`) and then we can collect metadata from all input zarr stores. The tricky bit is if two of the input stores have metadata under the same key - we'd need to pick one or throw an error and ask the user to decide. I'll draft an implementation. 

@aofei-liu I think [4a393ba](https://github.com/czbiohub-sf/iohub/pull/419/commits/4a393ba859024071151c40b6e1c3ffd3deb68749) does not resolve Ed's comment, I think we should roll back that commit.

aofei-liu:
@ieivanov sure, reverted. i didn't address sri's comment about the lists going out of sync yet (i have those commits locally, but didn't push since you asked for a revert), so remember to do that

the reason why i chose a map was because assemble in biahub already outputs a mapping of source -> dest that you can use to do the copies when assembling. I can try and adapt biahub code to work with whatever your implementation is

ieivanov:
I renamed the `copy_metadata_from` parameter to `metadata_sources` and allows users to pass in a list of stores from which extra zarr metadata will be copied. In case of duplicate entries, earlier sources in the list take precedence. @edyoshikun can you confirm this addresses the concatenate use case you brought up?

ieivanov (approved):

srivarra (approved):
Great, tyty for that fix.

aofei-liu:
Re: rolling back `copy_metadata_map` — I looked into this more and `ensure_unique_positions` in biahub defaults to `False` and isn't set to `True` in any real pipeline configs (only in tests), so the renamed-position-keys scenario isn't urgent. I think `metadata_sources: list[Path]` is the right call for now, and we can add explicit key remapping later if `ensure_unique_positions` sees real use.

One design question on the multi-source behavior: when two sources define the same zattrs key for a position, should the earlier source win (current implementation), or should we collect values from all sources into a list? e.g. if `source_a` has `extra_metadata: {"temp": 37}` and `source_b` has `extra_metadata: {"temp": 25}`:

- **First wins:** `extra_metadata: {"temp": 37}` (current — simple, predictable)
- **Append:** `extra_metadata: [{"temp": 37}, {"temp": 25}]` (preserves all provenance but changes the schema)

I think first-wins is fine for custom zattrs if they're typically uniform across a dataset, but wanted to flag it in case we are concatenating paths that have gone through different preprocessing and carry different custom metadata.

ieivanov:
>One design question on the multi-source behavior: when two sources define the same zattrs key for a position, should the earlier source win (current implementation), or should we collect values from all sources into a list? e.g. if source_a has extra_metadata: {"temp": 37} and source_b has extra_metadata: {"temp": 25}:

I think first wins is simplest and the right decision for now. I don't know of cases where we'd need metadata from different stores under the same key but with different values. Let's see if there is a strong use case for that and then we can amend. I'd argue a case like `extra_metadata: [{"temp": 37}, {"temp": 25}]` would be confusing and should be avoided at the start since we no longer know which one is true.

aofei-liu:
gotcha. i guess one copy is fine if we assume that data should be copied over from previous steps and there's no step that was run in isolation of other steps that would carry unique metadata not contained in the step downstream of that step.

speaking of that though, if we have multiple paths to concat, wouldn't the latest entry contain the most complete custom metadata? e.g.
```
 # from example_concatenate_settings_organelle_dynamics.yml
 concat_data_paths:
  - '../1-preprocess/label-free/1-stabilize/...'              # earliest
  - '../1-preprocess/label-free/2-virtual-stain/...'
  - '../1-preprocess/light-sheet/raw/2-register/...'
  - '../1-preprocess/light-sheet/deconvolved/2-stabilize-register/...'  # latest
```

so we'd want to keep the custom metadata from the last entry, not the first? I think you just need to drop the `k not in position.zattrs` guard to change the behavior from first wins to last wins

edyoshikun:
yeah I think you addressed my concerns @ieivanov. If we could as a start propagate metadata from waveorder reconstructions (i.e NA_det, illumination, pixel_sizexy and z, etc) that on it's own is already a plus. With `mantis-v2` datasets you wont have this issue during assembly since you aren't concatenating zarr stores.

edyoshikun (approved):

ieivanov:
Thanks @edyoshikun!

@aofei-liu yes, it's useful to work though a specific example

```
 # from example_concatenate_settings_organelle_dynamics.yml

 concat_data_paths:
# stabilized Phase3D, would contain metadata such as 'waveorder-reconstruct' and 'biahub-stabilize'
  - '../1-preprocess/label-free/1-stabilize/...'

# virtual staining of nuclei and membrane from stabilized Phase3D, would contain metadata such as 'waveorder-reconstruct', 'biahub-stabilize', and 'viscy-vs' (or whatever names we give them); 'waveorder-reconstruct' and 'biahub-stabilize' are exected to be the same as in the store above
  - '../1-preprocess/label-free/2-virtual-stain/...'

# raw (not deconvolved) deskewed fluorescence channels, would contain metadata such as 'biahub-deskew'
  - '../1-preprocess/light-sheet/raw/2-register/...'

# registered and deskewed fluorescence, would contain metadata such as 'biahub-deskew' and 'biahub-register'; 'biahub-deskew' is expected to the same as in the store above.
  - '../1-preprocess/light-sheet/deconvolved/2-stabilize-register/...'
```

In that case, when we concatenate, we should get the following metadata:
```
'waveorder-reconstruct': from ../1-preprocess/label-free/1-stabilize/
'biahub-stabilize': from ../1-preprocess/label-free/1-stabilize/
'viscy-vs': from ../1-preprocess/label-free/2-virtual-stain/
'biahub-deskew': from ../1-preprocess/light-sheet/raw/2-register/
'biahub-register': from ../1-preprocess/light-sheet/deconvolved/2-stabilize-register/
```

Does that make sense? I'll merge now. If we run into any issues we have the option to (1): not carry all metadata through intermediate steps, for example 2-virtual-stain could only have metadata on 'viscy-vs' and not 'waveorder-reconstruct' or 'biahub-stabilize', (2) we could rearrange with order of paths in `concat_data_paths` - there is currently nothing particular about the order, or (3) come back and fix this feature as needed.

Issue #266, opened 2025-01-21
Choose a faster default compressor


ziw-liu:
@shinzlet found that zstd compression that we use as default in iohub makes I/O much slower for zarr arrays without significant (<5%) disk space gain. The best choice seems to be among the on-the-wire flavor of compressors, which is unsurprising considering that zarr-python's default is lz4.

ziw-liu:
A more complete study for (slightly different) QPI data: https://arxiv.org/abs/2503.18037

shinzlet:
Interesting read! It seems pretty consistent with what I found in my testing too.

The biggest methodology difference between my work and the paper linked is probably chunk size - they use a 32x256x256 chunk for their compressor, I use chunks more like 100x2048x2048. That's a really massive difference and my testing shows that it definitely helps with bandwidth (and I imagine it must help with compression ratio).

When I get some free time, I would be really interested to do these benchmarks on windows / mac / linux / x64 / ARM platforms - when I started testing this months ago, I remember blosclz being a really strong performer compared to others, but all of our newer benchmarks seem to suggest that it's not very different than lz4 etc. I think the perf difference I saw might have been a result of doing those very first tests on my mac - perhaps the implementations perform differently on different builds / instruction sets.

PR #422, opened 2026-06-05
fix: merge extra_metadata in process_single_position instead of overwriting

aofei-liu:
## Summary

`process_single_position` unconditionally overwrites `extra_metadata` on the output position's zattrs, which causes two problems in multi-step pipelines:

1. **Clobber with `None`**: When no `extra_metadata` kwarg is passed, it writes `None`, destroying any metadata previously copied by `create_empty_plate`'s `metadata_sources` parameter (added in v0.3.7 / #419).

2. **Clobber with partial dict**: When `extra_metadata` is provided (e.g. `{"flat_field_correction": {...}}`), it replaces the entire existing dict rather than merging. In a pipeline like deskew → flat-field → reconstruct, each step's `extra_metadata` overwrites the previous step's keys.

### Fix

- When `extra_metadata` is **not provided**, existing metadata is left untouched.
- When `extra_metadata` is a **dict**, it is merged into any existing dict (new keys added, caller's keys take precedence).

### Context

Discovered while wiring `metadata_sources` through the [biahub PR cascade](https://github.com/czbiohub-sf/biahub/pull/255) (PRs #255–#266). The assembly/concatenation step copies metadata from 3 upstream plates via `metadata_sources`, then `process_single_position` immediately clobbers it with `None`. Upstream steps (flat-field, reconstruct) also lose prior steps' `extra_metadata` keys because the write is a full replacement.

A workaround is in place on the biahub side (read-back existing `extra_metadata` before calling `process_single_position`), but the root fix belongs here.

## Test plan

- [x] `test_process_single_position_preserves_existing_extra_metadata` — verifies omitting the kwarg preserves pre-existing metadata
- [x] `test_process_single_position_merges_extra_metadata` — verifies a new dict merges with pre-existing keys
- [x] Existing `test_process_single_position` still passes (backward compatible)

### DIFF
--- a/src/iohub/ngff/utils.py
+++ b/src/iohub/ngff/utils.py
@@ -408,6 +408,11 @@ def process_single_position(
         can be passed to be stored at a FOV level,
         e.g.,
         kwargs={"extra_metadata": {"Temperature": 37.5, "CO2_level": 0.5}}.
+        When ``extra_metadata`` is a dict it is **merged** into any
+        existing ``extra_metadata`` on the output position (so metadata
+        written by ``create_empty_plate``'s ``metadata_sources`` is
+        preserved).  When ``extra_metadata`` is not provided, existing
+        metadata is left untouched.
     """
     click.echo(f"Function to be applied: \t{func}")
     click.echo(f"Input data path:\t{input_position_path}")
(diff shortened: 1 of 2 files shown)


aofei-liu:
### Downstream cleanup (biahub) once this releases

There are workarounds in [`czbiohub-sf/biahub`](https://github.com/czbiohub-sf/biahub) that can be removed once this fix ships:

1. **`biahub/concatenate.py`** (branches: `refactor-assembly`, `refactor-qc`, `refactor-e2e`): reads back existing `extra_metadata` from the output position before calling `process_single_position` and passes it through as the `extra_metadata` kwarg — two call sites (debug executor path and submitit path). With this fix, those read-backs can be deleted since omitting the kwarg will preserve existing metadata.

2. **`biahub/flat_field_correction.py`**: currently passes `extra_metadata={"flat_field_correction": ...}` which replaces the upstream `{"deskew": ...}` key. With the merge semantics from this PR, the flat-field step's metadata will accumulate alongside deskew's — no code change needed in biahub, just verify after bumping iohub.

Tracking in biahub memory under "Pending: iohub extra_metadata merge fix (PR #422)".

ieivanov:
Closing in favor of #424

PR #425, opened 2026-06-12
fix(downsample): cover odd-sized axes when building pyramid levels

talonchandler:
**Summary**

When `compute_pyramid` (or any caller of `downsample_region` in the zarr-python
backend) operated on a source axis whose size was not an exact multiple of the
requested factor, the trailing half of that axis was silently discarded.
For an L0 shape with odd X=2055 and a factor-2 X downsample, L1 stored only
the leftmost 1028 columns of the Y-mean of L0 instead of a 2x2-pooled image.
At higher zoom napari then spread those leftmost columns across the full
world extent, producing visible X-stretching in OME-Zarr pyramids.

Surfaced while inspecting the daxi reconstruction sweep at
`/hpc/projects/waveorder/tile-stitch/runs/fused_sweep_iso_viz/` (L0 shape
`(2372, 2244, 2055)`, X is odd), generated by `iohub.compute_pyramid(dims={"y", "x"})`
via the `viz_pyramid.py` wrapper. Every store in the sweep had the same
left-half-only L1+ on the X axis.

**Root cause**

`ZarrPythonImplementation.downsample_region` recomputed a "local factor" per
region from `max(1, src_size // tgt_size)`. For the rightmost target region
on an odd-sized axis, `target_region_to_source` correctly clipped the source
slice to the array boundary, giving `src_size = 2055` and `tgt_size = 1028`.
Floor division then returned `1` instead of `2`, so `downsample_block`
returned a (1122, 2055) block. Zarr-python truncated that assignment to the
(1122, 1028) target region, keeping only the leftmost half of the source.

`downsample_block` itself also trimmed odd trailing blocks instead of covering
them, which independently disagrees with the ceiling-division shape that
`initialize_pyramid` allocates per level.

**Fix**

- `downsample_block` now edge-pads each axis up to the next multiple of its
  factor before pooling. Output shape is `ceil(s / f)`, matching the pyramid
  level shapes from `initialize_pyramid`.
- `downsample_region` no longer recomputes a per-region "local factor"; it
  passes the caller-supplied factors straight to `downsample_block`.

**Tests**

- `test_downsample_block_odd_axis`: pure-numpy regression for the trim vs.
  pad behaviour. A 21-wide ramp becomes an 11-wide averaged ramp whose last
  bin is 20.0 (edge-padded), not absent.
- `test_compute_pyramid_odd_axis`: end-to-end regression on a
  `(1, 1, 4, 6, 21)` zarr where Y is even and X is odd. L1's last X column
  equals 20.0 (correct edge-pad pool) instead of ~10 (left-half-only bug).

### DIFF
--- a/src/iohub/core/downsample.py
+++ b/src/iohub/core/downsample.py
@@ -8,7 +8,7 @@ import numpy as np
 
 
 def downsample_block(data: np.ndarray, factors: list[int], method: str) -> np.ndarray:
-    """Downsample an N-D numpy block by integer factors.
+    """Downsample an N-D numpy block by integer factors (1 = pass through).
 
     Axes whose size is not an exact multiple of the corresponding factor
     are edge-padded up to the next multiple before pooling, so the output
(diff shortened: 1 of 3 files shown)


talonchandler:
@srivarra in my trials this fixes the pyramid issue you were seeing. The X dimension had an odd length, which caused iohub's pyramid computation to misbehave.

srivarra:
@talonchandler I think this is good to be merged in, thanks for the fix!

srivarra (approved):

PR #427, opened 2026-06-22
refactor: Switching to Typer from Click

### LINKED ISSUE #426: Switch to Typer for CLI
In alignment with: https://github.com/czbiohub-sf/biahub/issues/253

srivarra:
Typer version `0.26` vendors Click instead of depending on it, so `OptionEatAll` needed some adjustments.

- [x] Might be useful to expose the `OptionEatAll` for downstream libraries (i.e. `biahub`) too...


ieivanov:
Looks very nice, thanks. 

Screenshot 2026-06-25 at 2 48 41 PM

I agree it would be nice to expose `OptionEatAll` for other libraries so we don't maintain this code in multiple places.

The CLI interaction seems noticeably slower however. Did you have the same experience? I'd expect it to be a bit slower given the rich formatting, but this may be borderline. Is there something we can do about it?

srivarra:
@ieivanov 
> The CLI interaction seems noticeably slower however. Did you have the same experience?

Yeah it's a known issue with building cli's with Typer/Python, there's a handful of discussions about this topic. Importing rich + others seems to cause the slowdown, not rich doing the formatting. Lemme take a look and try some stuff and get back to you.

srivarra:
@ieivanov 

The main slowdown is when invoking the CLI, we import every package, and I think typer+rich made it slower enough to notice.

I tried out [`lazy_loader`](https://scientific-python.org/specs/spec-0001/) and made it snappier.

| command | Click (main) | Typer | Typer + lazy_loader |
|---|--:|--:|--:|
| `python -c pass` (floor) | 18 ms | 20 ms | 18 ms |
| `iohub --version` | 536 ms | 598 ms | **168 ms** |
| `iohub -h` | 551 ms | 629 ms | **210 ms** |

ieivanov (approved):
Looks nice. It's worth getting review from someone else too before you merge

ieivanov:
I'd say let's merge this PR now, it's important we don't let PRs go state. We can fix any issues that we haven't caught here later.

PR #438, opened 2026-07-10
fix(ngff): avoid member listing when indexing container nodes

edyoshikun:
## Summary

`NGFFNode.__getitem__` used `if not znode:` on the value returned by `zarr.Group.get` (container branch, reached for `Plate`→`Well` and `Well`→`Position`). `zarr.Group` defines `__len__` but **no** `__bool__`, so the truthiness test silently fell back to `__len__` → `nmembers()` — a full member listing — on **every** container open. It was also a latent bug: an existing but empty group has zero members, so `not znode` was `True` and raised a spurious `KeyError` for a group that exists (inconsistent with `__contains__`, which uses member names).

`.get` already returns `None` for a missing key, so `if znode is None:` is both correct and cheap.

```python
znode = self.zgroup.get(key)
-if not znode:
+if znode is None:
    raise KeyError(key)
```

## Why the one-liner matters (mechanism)

`znode` is a `zarr.Group`, so evaluating `not znode` is **not** a cheap check — with no `__bool__`, Python falls back to `len(znode)` → `Group.__len__` → `nmembers()`, which lists the group's members over storage. Measured on a real group (6 members):

| expression | cost/call |
|---|---|
| `znode is None` (this PR) | **0.036 µs** (pointer identity) |
| `not znode` (before) | **~7,100 µs** (triggers `nmembers()` member listing) |

The speedup isn't a faster comparison — it's *avoiding an entire member-listing I/O op* that the truthiness check was silently performing on every open.

## Impact

Profiling a real HCS plate (OME-Zarr v0.5, position `zarr.json` ~107 KB), measuring `plate[fov_name]` open cost, 200 opens each:

| page cache | before (`if not znode`) | after (`if znode is None`) |
|---|---|---|
| **warm** (median) | ~2.95 ms | **~1.72 ms** (~40% faster) |
| **cold, NFS** (median) | ~6.6–6.9 ms | ~3.5–6.4 ms |

The `Group.__len__`/`nmembers()` call is eliminated entirely (verified by tracing: 0 calls during a position open). Warm-cache is the clean headline; cold NFS reads are noisier (network round-trips inflate the floor for both), but the fix still helps. The path is backend-independent — `ZarrsPythonImplementation` delegates group operations to zarr-python, so the default `zarrs-python` backend is affected too.

## Behavior change

An existing but empty group (e.g. a well with no positions) is now **returned** instead of raising `KeyError`. This aligns `node[key]` with `key in node`, which already reported such groups as present. A genuinely missing key still raises `KeyError`.

Verified safe at every container level: constructing a node on an empty group never raises a non-`KeyError` — `Row`/`Well`/`Plate` `_parse_meta` only `_warn_invalid_meta()` (logs), and `Position._parse_meta` catches `ValidationError` and warns. Multi-level descent (`"A/1/0"`) is unaffected.

## Test

`test_getitem_returns_existing_empty_group` reuses the existing `_temp_ome_zarr_plate` helper and the public `create_well` API: asserts an existing empty well is returned as a `Well`, is empty (`list(node) == []`) but usable, and that a missing key still raises `KeyError`. Verified it **passes with the fix and fails (`KeyError`) without it**.

### DIFF
--- a/src/iohub/ngff/nodes.py
+++ b/src/iohub/ngff/nodes.py
@@ -284,10 +284,7 @@ class NGFFNode:
             return item_type.from_handle(handle, self._impl)
         else:
             znode = self.zgroup.get(key)
-            # Use ``is None`` rather than truthiness: ``zarr.Group`` has no
-            # ``__bool__``, so ``not znode`` would fall back to ``__len__`` ->
-            # ``nmembers()``, a full member-listing that dominates the
-            # position-open cost. ``.get`` already returns ``None`` when absent.
+            # Not ``not znode``: zarr Group has no __bool__, so truthiness calls __len__ -> nmembers().
             if znode is None:
                 raise KeyError(key)
             return item_type(group=znode, parse_meta=True, **self._child_attrs)
--- a/src/iohub/ngff/nodes.py
+++ b/src/iohub/ngff/nodes.py
@@ -284,7 +284,11 @@ class NGFFNode:
             return item_type.from_handle(handle, self._impl)
         else:
             znode = self.zgroup.get(key)
-            if not znode:
+            # Use ``is None`` rather than truthiness: ``zarr.Group`` has no
+            # ``__bool__``, so ``not znode`` would fall back to ``__len__`` ->
+            # ``nmembers()``, a full member-listing that dominates the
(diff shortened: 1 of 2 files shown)


srivarra (approved):
Cool optimization, thanks!

PR #444, opened 2026-07-24
feat: add consolidated metadata support

### LINKED ISSUE #435: Consolidated Metadata
Even though [consolidated metadata](https://zarr.readthedocs.io/en/stable/user-guide/consolidated_metadata/) is still experimental for Zarr V3, it would be nice to be able to have the option to write all store metadata at the root group of the Zarr store.

srivarra:
Adds consolidated metadata to `NGFFNode`. Therefore at any hierarchy of an OME-Zarr you can consolidate metadata.

### HCS

```python

  with open_ome_zarr("plate.zarr", layout="hcs", mode="r+") as plate:
      plate.consolidate_metadata()
```

### Well

```python
  with open_ome_zarr("plate.zarr", layout="hcs", mode="r+") as plate:
      well = plate["A/1"]
      well.consolidate_metadata()
```

### DIFF
--- a/src/iohub/ngff/nodes.py
+++ b/src/iohub/ngff/nodes.py
@@ -418,6 +418,23 @@ class NGFFNode:
         """Close Zarr store."""
         self._impl.close(self._group)
 
+    def consolidate_metadata(self) -> None:
+        """Consolidate this node's descendant metadata into its own metadata file.
+
+        Speeds up opening large hierarchies (e.g. HCS plates) since readers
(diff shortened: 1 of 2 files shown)


ieivanov:
@srivarra could you give us more background on what consolidated metadata is, how we use it currently, or how we could use it in the future?

srivarra:
@ieivanov 

One of the main uses would be that we know every FOV in a plate without needing to traverse the directory tree, it's just in the base store's json file.

PR #454, opened 2026-08-04
perf(ngff): let create_empty_plate filter and batch copied position metadata

ieivanov:
## Problem

`create_empty_plate(metadata_sources=...)` copies **every** non-OME zattrs key from the source position, **one assignment at a time**. Both halves hurt when a source plate carries a large instrument-written blob.

**Nothing filters the copy.** A mantis acquisition writes `ome_writers.frame_metadata` — one record per raw Z-frame:

```
214,668 entries  ==  T*C*Z of the raw store (67 x 3 x 1068)
18.5 MB compact JSON, 39 MB as written
```

That is acquisition bookkeeping, and it propagates into every derived plate. It is also *stale* on arrival: after deskewing, the destination has Z=86, so the copied `storage_index` entries describe a frame grid that no longer exists.

**The write is quadratic.** Assigning to `position.zattrs` serializes the whole group metadata document, and the `k not in position.zattrs` guard deserializes it. Per key, over a 39 MB document.

Measured on a 72-position plate, in the `--init` step whose only job is to create an empty plate:

| step | wall | read | written |
|---|---|---|---|
| flat-field | 3m 09s | 1.0 GB | 4 GB |
| deskew | 6m 14s | 2.3 GB | 4 GB |
| reconstruct | 7m 28s | 2.1 GB | 6 GB |
| concatenate (3 sources) | 14m 38s | 16.1 GB | 10 GB |
| virtual-stain — *no `metadata_sources`* | **37s** | 273 MB | **632 KB** |

The last row is the control: the one step that doesn't pass `metadata_sources` is the one that's fast.

## Change

- Add `metadata_keys`: `fnmatch` patterns selecting which keys a source may contribute. Default `None` keeps today's copy-every-non-OME-key behaviour, so this is backwards compatible.
- Collect across all sources first, then write once via `Attributes.put`.

Worth noting for reviewers: `Attributes.update` is the `MutableMapping` default and delegates to per-key `__setitem__`, so it would *not* have fixed the amplification. `put` is the single-write path, hence the explicit `{**existing, **collected}` merge.

## Result

Downstream in biahub, passing `metadata_keys=("biahub-*", "waveorder", "cytoland")`:

- per-position `zarr.json`: **39 MB -> 2.7 KB**
- `deskew --init` over 66 positions: **6m 14s -> 49s** (most of the remainder is interpreter startup)
- array metadata and the OME block are byte-identical to the store built by the current code

## Tests

Three added:

- `metadata_keys` filters by pattern
- `metadata_keys=None` still copies everything
- the copy is a single write — asserts the write count does **not** grow with the number of keys copied, rather than pinning an exact count. Verified it fails against the per-key implementation.

`tests/ngff/test_ngff_utils.py`: 36 passed.

### DIFF
--- a/src/iohub/ngff/utils.py
+++ b/src/iohub/ngff/utils.py
@@ -228,15 +228,17 @@ def create_empty_plate(
             existing = dict(position.zattrs)
             collected: dict[str, Any] = {}
             for source in metadata_sources:
+                # Only the open is guarded: a source that lacks this position
+                # is skipped, but a failure while reading one is a real error.
                 try:
                     src_pos = open_ome_zarr(source / position_key_string, layout="fov", mode="r")
                 except FileNotFoundError:
                     continue
-                src_attrs = dict(src_pos.zattrs)
+                with src_pos:
+                    src_attrs = dict(src_pos.zattrs)
                 for k in _selected_metadata_keys(src_attrs, metadata_keys):
                     if k not in existing and k not in collected:
                         collected[k] = src_attrs[k]
-                src_pos.close()
             if collected:
                 position.zattrs.put({**existing, **collected})
         else:
--- a/src/iohub/ngff/utils.py
+++ b/src/iohub/ngff/utils.py
@@ -30,13 +30,20 @@ _OME_KEYS = {"ome", "multiscales", "omero", "labels", "version"}
 
 def _selected_metadata_keys(
     source_attrs: Mapping[str, Any],
-    metadata_keys: Iterable[str] | None,
+    metadata_keys: str | Iterable[str] | None,
 ) -> list[str]:
     """Names of the source zattrs to copy to a destination position.
 
     OME-owned keys are always excluded. When ``metadata_keys`` is None every
     remaining key is selected; otherwise a key is selected only if it matches
     at least one of the ``fnmatch`` patterns.
+
+    Patterns are shell globs rather than regexes: the strings callers actually
+    write, like ``"biahub-*"``, are also valid regexes that mean something
+    else, so reading them as regexes would silently select nothing instead of
+    raising. The iterable already supplies the alternation a regex would add.
+    ``fnmatchcase`` keeps matching case-sensitive on every platform, matching
+    how zattrs keys compare.
     """
     candidates = (k for k in source_attrs if k not in _OME_KEYS)
     if metadata_keys is None:
@@ -59,7 +66,7 @@ def create_empty_plate(
     scale: tuple[float, ...] = (1, 1, 1, 1, 1),
     dtype: DTypeLike = np.float32,
     metadata_sources: Path | str | list[Path | str] | None = None,
-    metadata_keys: Iterable[str] | None = None,
+    metadata_keys: str | Iterable[str] | None = None,
 ) -> None:
     """
     Create a new HCS Plate in OME-Zarr format if the plate does not exist.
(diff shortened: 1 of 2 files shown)


src/iohub/ngff/utils.py
```diff
@@ -6,8 +6,9 @@
 import os
 import warnings
 from collections import defaultdict
-from collections.abc import Callable, Mapping, Sequence
+from collections.abc import Callable, Iterable, Mapping, Sequence
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor, as_completed
+from fnmatch import fnmatchcase
```
srivarra, line 11: [thought(non-blocking)](https://pullpo.io/cc?l=thought&d=non-blocking) 
didn't even know this was part of the standard library for python, cool. Why not use plain regex though?

src/iohub/ngff/utils.py
```diff
@@ -78,15 +100,24 @@ def create_empty_plate(
...
         precedence over later ones). Coordinate transforms, axis definitions,
         and label references are **not** copied.
         Defaults to None (no metadata copy).
+    metadata_keys : iterable of str, optional
+        ``fnmatch`` patterns narrowing which zattrs keys ``metadata_sources``
+        may contribute, e.g. ``{"provenance-*", "acquisition"}``. A key is
+        copied only if it matches at least one pattern; OME-owned keys are
+        excluded either way. This only filters the sources, so on its own it
+        copies nothing — it has no effect unless ``metadata_sources`` is set.
+        Defaults to None (copy every non-OME key the sources provide).
```
srivarra: [nitpick](https://pullpo.io/cc?l=nitpick) 
Can we raise a error if metadata keys is passed without metadata sources?

srivarra (changes requested):
One small request, but otherwise looks good. the perf improvement is surprisingly quite a lot. tyvm!

src/iohub/ngff/utils.py
ieivanov: Tried it both ways, and the deciding factor is that the patterns we actually intend people to write are valid in *both* languages while meaning different things.

The usage pattern this is for is "take the provenance families, leave the instrument blob" — downstream in biahub that's literally `metadata_keys=("biahub-*", "waveorder", "cytoland")`. Run that exact tuple against the keys on a real derived plate (`biahub-flat_field`, `biahub-deskew`, `biahub-reconstruct`, `biahub-concatenate`, `waveorder`, `cytoland`, `ome_writers.frame_metadata`, `acquisition`, `frame_log`):

| pattern | `fnmatchcase` | as regex, `fullmatch` | as regex, `search` |
|---|---|---|---|
| `("biahub-*", "waveorder", "cytoland")` | the 4 `biahub-` keys + `waveorder` + `cytoland` | **just `waveorder`, `cytoland`** | the 4 + 2 ✅ |
| `("biahub-*",)` | the 4 `biahub-` keys | **`[]`** | the 4 ✅ |
| `("biahub-[dr]*",)` | `biahub-deskew`, `biahub-reconstruct` | **`[]`** | **all 4** |

`biahub-*` as a regex means "`biahub`, then zero or more hyphens" — so under regex it doesn't fail, it silently selects the wrong set. With `fullmatch` you get nothing copied; with `search` a glob character class quietly over-copies. Either way the symptom is wrong provenance metadata on the derived plate, found much later by whoever reads it. Globs also have no invalid-pattern case at all (a stray `[` is just a literal), so there's no `re.error` path to design a message for.

The main thing regex would buy us — alternation — is already in the API: the iterable *is* the OR, so `("biahub-deskew", "biahub-reconstruct")` covers `biahub-(deskew|reconstruct)`.

The one thing regex can genuinely express and globs can't is a negative lookahead — `(?!ome_writers\.).*`, i.e. "everything except the blob". That's a fair want given the blob is the whole point of this PR, but I'd rather add an explicit exclusion parameter later than change the meaning of this one; it composes fine with globs.

Kept `fnmatchcase` (not `fnmatch`) so matching stays case-sensitive on every platform, the way zattrs keys compare. Reasoning is now in the `_selected_metadata_keys` docstring, and the parameter docs say "shell-glob patterns (`fnmatch`, not regex)".

src/iohub/ngff/utils.py
ieivanov: Done in c8a14e9. It raises before the plate is opened, so a misuse leaves nothing behind:

```python
if metadata_sources is None:
    if metadata_keys is not None:
        raise ValueError(
            "metadata_keys filters what metadata_sources contributes, "
            "so it selects nothing on its own. Pass metadata_sources, "
            "or drop metadata_keys."
        )
```

Also added a `Raises` section to the docstring and a test that asserts the destination store is never created.

One judgment call worth flagging: this fires for `metadata_sources=None`, not for an explicit empty list. A caller that computes `sources = [...]` and passes a fixed `metadata_keys` alongside it shouldn't blow up when the list legitimately comes out empty. Happy to make it strict about both if you'd rather.

src/iohub/ngff/utils.py
srivarra: Cool makes sense to me!

src/iohub/ngff/utils.py
srivarra: I think `None` is good enough, not sure anyone will pass an empty list tbh.

srivarra (approved):

PR #460, opened 2026-08-14
Write a gapped time selection as contiguous runs

ieivanov:
## The failure

Concatenating a `(67, 6, 86, 1664, 1193)` float32 plate with
`shards_ratio: [5, 1, 6, 7, 5]` died with:

```
numpy._core._exceptions._ArrayMemoryError: Unable to allocate 651. TiB for an
array with shape (1048576, 86, 1664, 1193) and data type float32
```

## Why

Sharding along T makes `process_single_position` batch a whole shard's worth of
timepoints into one write. `apply_transform_to_tczyx_and_save` then drops the
timepoints whose input is all zeros or NaNs — here `t=1` and `t=2` of the
`t=0..4` batch — and writes the survivors, `[0, 3, 4]`, in a single orthogonal
write. A caller asking for non-consecutive `output_time_indices` outright
produces the same shape of selection.

A gapped selection is not expressible as a slice, and a sharded write needs
one:

1. the zarrs pipeline rejects it — `DiscontiguousArrayError: [3 1]`, which is
   exactly `np.diff([0, 3, 4])`;
2. it falls back to zarr-python's partial-shard encode path, which reaches
   `chunk_value = value[out_selection]` with `out_selection` carrying
   coordinates from the *shard's* index space rather than the value buffer's.
   On a small array that is an `IndexError`; on a real one numpy is asked for
   the whole index space.

## The change

Split the write at every gap and write each run of consecutive timepoints
separately. Each run is sliceable, which both pipelines handle.

- The no-gap case — every batch where nothing was dropped — stays a single
  write, so the common path is unchanged.
- Peak memory is unchanged: the same buffers are written, in more than one
  call.
- `WriteUnit.begin()`/`complete()` still bracket the whole set of runs, so a
  kill mid-way leaves the in-flight marker and the unit is recomputed. Repair
  and resume behave as before.

## Tests

Two regression tests, both of which fail on `main` with the production
signature (`DiscontiguousArrayError` → `IndexError`) and pass here:

- `test_process_single_position_with_zero_timepoints_inside_a_t_shard` — blank
  `t=1,2` inside a 5-deep T shard, asserting the skipped timepoints stay at the
  fill value and the rest are transformed.
- `test_apply_transform_to_tczyx_with_gapped_output_time_indices` — a
  caller-requested gap within one T shard.

Plus a table-driven unit test for the `_contiguous_runs` helper.

`tests/ngff/test_ngff_utils.py` is otherwise green. The one failure,
`test_apply_transform_to_tczyx_and_save`, reproduces identically on unmodified
`main`: a zarrs `incompatible offset` error for a shard spanning multiple
*channels*, which is unrelated and not touched here.

### DIFF
--- a/src/iohub/ngff/utils.py
+++ b/src/iohub/ngff/utils.py
@@ -316,22 +316,55 @@ def _echo_finished(
         click.echo(f"Finished writing t={time_index}, c={channel_index}")
 
 
+def _contiguous_runs(indices: Sequence[int]) -> list[slice]:
+    """Split ``indices`` into maximal runs of consecutive values.
+
+    Returns slices *into* ``indices`` rather than the values themselves, so one
+    call carves up both an index list and the data written at those indices.
+    """
+    runs: list[slice] = []
+    start = 0
+    for position in range(1, len(indices) + 1):
+        if position == len(indices) or indices[position] != indices[position - 1] + 1:
+            runs.append(slice(start, position))
+            start = position
+    return runs
+
+
 def _save_transformed(
-    transformed: NDArray | list[NDArray] | None,
+    transformed: list[NDArray],
     output_position_path: Path,
     output_channel_indices: list[int] | slice,
-    output_time_indices: int | list[int],
+    output_time_indices: list[int],
     write_unit: WriteUnit | None = None,
 ) -> None:
+    """Write the transformed timepoints, one run of consecutive timepoints at a time.
+
+    The write has to be split because a *gapped* time selection is not
+    expressible as a slice, and a sharded write cannot be done without one:
+    the zarrs (Rust) codec pipeline rejects the selection with
+    ``DiscontiguousArrayError`` and falls back to a zarr-python path that
+    mismaps the value buffer onto the shard — indexing it with coordinates from
+    the shard's own space, which raises ``IndexError`` or, on a large array,
+    asks numpy for a preposterous allocation.
+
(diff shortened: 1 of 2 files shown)


srivarra:
@ieivanov lemme make sure I'm understanding this one correctly. 

If we are concatenating nonconsecutive timepoints that are on the same (?) shard this causes problems. So instead we are breaking apart the timepoints into set of consecutive timepoints such that we can use `oindex` with for writing. Anything I've forgotten?

ieivanov:
@srivarra that's largely correct.

The problem arises when we tried to put several timepoints into the same shard, but some of these timepoints are missing data (e.g. all zeros). We haven't done that much before, so we haven't run into that issue yet. I'm now trying to make sure our assembled zarr stores have shards of few GB in size which requires lumping several timepoints into one shard.

Here is an example. Let's say we have a dataset with `t=0..4` and `t=1` and `t=2` are all zeros (fill value). We'd like to store these data in the same shard. Previously this used to throw errors because zarrs `make_slice_selection` doesn't like the `[0, 3, 4]` indices we are trying to pass. So now we do two writes into that chunk with `[slice(0, 1), slice(3, 5)]`. This is two read-modify-write operations rather than just one write for non-gapped shards, but that acceptable. 

Let me know if that makes sense and if the implementation matches my description.

srivarra:
@ieivanov Ahhh, yeah that makes sense!

srivarra (approved):

